PR #21, opened 2017-12-19
modules/drivers: MCP230XX refactor/redesign

rwaldron:
- [MCP23008 Example Video](https://www.youtube.com/watch?v=3jL3YbmzFH4)
- [MCP23017 Example Video](https://www.youtube.com/watch?v=11KogmkKdG8)

Summary: 

- `MCP23008` and `MCP23017` are subclasses of `Expander`
- `Expander` is a subclass of `SMBus`
  - `Expander` instances can read & write the entire GPIO bank
- `Expander` instances are iterable
  - Entries are `Pin` instance objects
    - `Pin` instances can read, write or set the mode of a single pin

More explanations are included in the docs


phoddie:
Overall this looks really good. I learned a few nice techniques from the code. I have a few suggestions. They are based on reading the code, as I don't have either if the expander parts on hand to try. Apologies for any errors.

### Device specific constants
I suggest moving the `IODIR`, `GPPU`, and `GPIO` registers to the prototype, together with `pins`. That eliminates the `reg` object which saves some RAM and moves more of the initialization work to build time. Here's the original `MCP23008`

	class MCP23008 extends Expander {
	  constructor(dictionary = {}) {
	    super(
	      Object.assign(dictionary, {
	        pins: 8,
	        reg: {
	          IODIR: 0x00,
	          GPPU: 0x06,
	          GPIO: 0x09,
	        }
	      })
	    );
	  }
	}

Here's the revised version:

	class MCP23008 extends Expander {
	}
	MCP23008.prototype.pins = 8;
	MCP23008.prototype.IODIR = 0x00;
	MCP23008.prototype.GPPU = 0x06;
	MCP23008.prototype.GPIO = 0x09;
	Object.freeze(MCP23008.prototype);

We usually freeze the prototypes of classes. Doing so saves an entry in the alias table (4 bytes of RAM!) and ensures the prototype won't change unexpectedly.

With the above change, in the constructor, access to `pins` moves from `dictionary` to `this`, so this...

	class Expander extends SMBus {
	  constructor(dictionary) {
	
	    const offset = (dictionary.pins >> 3) - 1;

...becomes....

	    const offset = (this.pins >> 3) - 1;

Uses of `IODIR`, `GPPU`, and `GPIO` adjust similarly. For example, here...

	bankWrite(state) {
		const { IODIR, GPIO } = this.reg;

		if (this.offset) {
			// Read IODIR state
			let iodir = this.readWord(IODIR);

...can eliminate the overhead of destructing and retrieve `IODIR` from `this` directly:

	let iodir = this.readWord(this.IODIR);
	
One small question about naming, which is admittedly only a question of style. Since `bankWrite` and `bankRead` are on the `Expander` itself, could they simply be named `write` and `read`?

### Pin array
Using the `Expander` instance as an `Array` to store the pins is elegant, and allows the pins instances to emulate the interface of Digital pins exactly.

I'm not sure the implementation of the `Symbol.iterator` symbol in the `Expander` is necessary. Where it could be used in the example, it isn't. The overhead is low, so code footprint isn't a concern, so I ask only in the interest of microcontroller minimalism. :)

### I2C
I should explain a bit about the intent beyond dictionary passed to the `I2C` constructor. Like `Digital`, `I2C` has a `port` string as part of dictionary to specify the hardware connection. The current device implementations all use `port` of  `NULL` implicitly so this isn't at all obvious. The `port` is needed, for example, on devices with several I2C buses, which may only be addressable by name (I believe Zephyr works this way) or by name in combination with pin numbers.

The consequence of that is that the `Expander` constructor should pass the ``dictionary` directly to `super` (with the I2C `address` applied, when needed) rather than creating a new `dictionary` with the expected fields. That will allow all relevant properties, current and future, to propagate.

There's no need to provide default pins for `scl` and `sda`. When they are absent (and no `port` is provided), the device's default I2C connection is used. For example, on ESP8266 the default pins are are defined in [`manifest_base.json`](https://github.com/Moddable-OpenSource/moddable/blob/7e6719730c73539c858f6a49440787a12fa7aa19/examples/manifest_base.json#L65) and applied in [`modI2c.c`](https://github.com/Moddable-OpenSource/moddable/blob/7e6719730c73539c858f6a49440787a12fa7aa19/modules/pins/i2c/esp/modI2C.c#L68).

FWIW, the default I2C connection is there as a convenience. It helps avoid having code, especially sample code, have dependencies on a particular host microcontroller by eliminating any pin specifications in the common default case.

rwaldron:
Re: moving things to MCP23008.prototype & MCP23017.prototype, that sounds good. 

> Since bankWrite and bankRead are on the Expander itself, could they simply be named write and read?

I wish! I actually forgot to follow up on this... I think there is a bug in class inheritance. I have just not filed an issue: https://github.com/Moddable-OpenSource/moddable/issues/22


> I'm not sure the implementation of the Symbol.iterator symbol in the Expander is necessary. Where it could be used in the example, it isn't.

I struggled with "lack of an example" support piece as well, but for the examples I wanted to preserve as much of the existing example code as I could. My thinking was: if you're +1 to the Symbol.iterator idea, then I would write some interesting examples for the documentation. I've decided to remove it because it can always be added later if there is user interest/demand. 


> I2C...

Got it—I will make the appropriate updates.

rwaldron:
@phoddie pushed updates!

rwaldron:
I don't know why the PR is not updating with my new commit :| https://github.com/rwaldron/moddable/commit/cf5c141bf0bcbaf96fbc5ffdba2f71878389c48a

rwaldron:
ok... it looks like my fork is no longer a valid fork of this repo :(

rwaldron:
https://github.com/Moddable-OpenSource/moddable/pull/23

rwaldron:
Here: https://github.com/Moddable-OpenSource/moddable/pull/23/commits/cf5c141bf0bcbaf96fbc5ffdba2f71878389c48a I've done my best to call out the relevant sections from the review above

PR #32, opened 2018-01-28
doc: fix tmp36 conversion algorithm

fivdi:
Currently the TMP36 conversion formula is documented in pins.md as:
```js
let temperature = (Analog.read(0) / 1024) * 330;
```
The maximum value returned by `Analog.read(0)` is 1023 so it should be divide by 1023 rather than by 1024. The Adafruit tutorial linked to is also incorrect here. The TPM36 also has a 500mV or 50°C offset that needs to be taken into account. The correct conversion formula is therefore:
```js
let temperature = (Analog.read(0) / 1023) * 330 - 50;
```


fivdi:
When a NodeMCU board is used pin A0 on the header isn't directly connected to the ESP8266. Pin A0 is the input voltage of a voltage divider and the output voltage of the voltage divider is directly connected to the ESP8266. See [here](https://github.com/nodemcu/nodemcu-devkit-v1.0/blob/master/NODEMCU_DEVKIT_V1.0.PDF) for details. If a board doesn't have a voltage divider for the analog input the formula to use is:
```js
let temperature = (Analog.read(0) / 1023) * 100 - 50;
```
I'm not sure if this is worth mentioning as the default board for the documentation is the NodeMCU if I'm not mistaken.

phoddie:
Edit looks good. We'll merge that this week.  Since the docs specially call out 8266, it makes sense to also add your note about the voltage divider.

fivdi:
How about adding the below text?

Caution: The output voltage range of the TMP36 temperature sensor is
0.1V to 2.0V and the input voltage range of the analog to digital converter on
the ESP8266 is 0V to 1.0V. To avoid damaging the ESP8266 a voltage divider
should be used to reduce the magnitude of the TMP36 output voltage. The
ESP8266 NodeMCU board has a voltage divider for this purpose. Other boards
may not have a voltage divider.

rwaldron:
@fivdi Nice! I'm excited to see you here :D

fivdi:
@rwaldron probably not as excited as I am to see you here :smile:

phoddie:
@fivdi Additional text looks good. Want to add it to the pull request?

fivdi:
@phoddie 

> Want to add it to the pull request?

Will do. At about 20:00 CET.

PR #37, opened 2018-02-04
modules/pins/analog: add adc for the esp32

fivdi:
This PR adds ADC for the ESP32. The ESP32 has two ADC peripherals, ADC1 and ADC2. This PR only adds support for ADC on ADC1 but all eight channels are supported. Support for ADC2 was not added as ADC2 can only be used when Wi-Fi is not used. For additional info see http://esp-idf.readthedocs.io/en/latest/api-reference/peripherals/adc.html#overview.

Note that there are still a number of issues with ADC on the ESP32.

ADC is non-linear and V-REF varies from chip to chip. See https://github.com/espressif/esp-idf/issues/164 and in particular https://github.com/espressif/esp-idf/issues/164#issuecomment-323994850 for more information.

ADC noise and cross coupling. See https://github.com/espressif/esp-idf/issues/1517 for more information.

This PR calls `adc1_to_voltage` which uses look up tables to avoid the non-linear issue. The millivolt value returned by `adc1_to_voltage` is then converted to a value between 0 and 1023 which should be linear.


phoddie:
Very nice. I tried to implement this a while back but the ADC issues on ESP32 discouraged me so I tossed the code. This looks better I'll review Monday.

fivdi:
> ADC issues on ESP32 discouraged me

They discouraged me too. ADC on the ESP32 is suboptimal at the moment. This PR tries to make the best of the current situation.

phoddie:
Code looks as good as it can. We'll merge this. Also, will add a manifest for Analog to be consistent with i2c, spi, and digital.

PR #73, opened 2018-07-12
Preliminary version of hardware Serial communication module for esp32

diogoviannaaraujo:
Hi,
After 2 days reading the documentation, other modules and lots of compile error I think I did get a funcional Serial module for the Esp32. Ive based on the i2c module and the esp idf examples.

This is preliminary work and I have no prior C experience, so this probably is nowhere near stable.

I was able to get a loopback response and do some initial talk with a SIMCom SIM800L GSM module. I do get some trash chars sometimes when talking with the GSM module, but don't know if its the module or something in the code.

Im posting this seeking for review on my code and some guidance.
My main difficulties are about the buffer, max message sizes and converting those buffer arrays into strings.
Since im usually programming on high level high memory environments any feedback in how to correct handle the data for this module is really appreciated.

Thanks in advance,
Diogo


diogoviannaaraujo:
Hi is hardware uart a desired function for the project?

Moddable-OpenSource:
Diogo, I had sent this to your email, but perhaps it didn't arrive:

We will need a signed Contributor license before we can accept the
pull request.

It can be found at:

https://github.com/Moddable-OpenSource/moddable/blob/public/licenses/Moddable%20CLA%20-%20individual.pdf

PR #78, opened 2018-08-03
WIP: Optimize Linux make files for finer grained control and more concurrency

creationix:
These are just some quick tweaks to the Makefiles to make my build much faster for building host tools.

Using `$(MAKE)` instead of `make` passes through options like `-j8` to child makefiles.

Splitting out the big `debug` and `release` targets into multiple targets allows concurrency as well fine-grained control.  For example, I only need `xsc-release` and `xsl-release` for my app.  I can now build those in 15s from a cold clone (It took minutes before)


creationix:
If you like this, I can apply the same changes to the mac files.

phoddie:
Spiffy. You probably eventually will want to build xsbug so you can debug. ;)

To accept the pull, we need a [CLA](https://github.com/Moddable-OpenSource/moddable/blob/public/licenses/Moddable%20CLA%20-%20individual.pdf).

Yes, a Mac version would be great.

PR #118, opened 2019-01-30
Add basic wasm build for xsc/xsl

FWeinb:
As discussed in #116 I created a makefile for building xsc/xsl using emscripten. I noticed that in the end the changes needed where minimal. I did change the name from `js` to `wasm` as discussed there. 

Currently the build for `xsl` depends on the presence of the `mac` build artifact `xsc`, this needs to change ([see here](https://github.com/Moddable-OpenSource/moddable/compare/public...FWeinb:wasm-support?expand=1#diff-44746cc29df986c47ec10001bd53486eR34)).  

To test these modules one can use the little tool I showcases in #116. 

@phoddie 
If you have any questions, feel free to ask.

[Edit] 
I worked on this some more. I improved the build script by including a working release build (which will take some time to create, especially for xsl)
To get the most of the compiled scripts I choose to compile them to be used inside a web worker. Because running on the main thread isn't really feasible in the long run.

I updated the little WebIDE and got the gzipped transfer size down to ~350Kb. You can test it [here](https://distracted-hoover-d4cde4.netlify.com/). 
> Keep in mind to allow insecure scripts after hitting the `install` button because of mixed-content police regarding https/http resources in the browser.  

One problem I need to figure out is how to share the two emscripten filesystems between xsc and xsl. Currently I am just copying the file between these two but that is not going to work if there are multiple files involved and we want to support incremental building etc. But I am fairly certain that there is a way to share the filesystems.

PR #132, opened 2019-02-24
Wasm support

FWeinb:
As requested by #118 I merged my branch with the latest changes in public and added an initial draft for `xscl` (combining `xsc` and `xsl` into on wasm build artifact).  

This PR is quite a mess because of the force-push that closed the last PR of mine. I did not had the time to to a proper rebase. 

The [WebIDE](https://distracted-hoover-d4cde4.netlify.com/) is running a build of `xscl` with all the changes included here. 

If the functions to build/link would be abstracted into one place and then used in `xsc`/`xsl`/`xscl` the maintenance overhead would be minimal. For the wasm build the debugger can be excluded as well. 

The merge of xsc/xsl into xscl is very primitive I just copy paste both tools and removed nearly all command line options. In the long run there is a need for some options (e.g. -debug) to be added back to the xscl.


phoddie:
Thanks again for resubmitting this PR. I've merged the changes. They will be pushed to this GitHub repository in the next day. When that happens, this PR will be automatically closed. (We have made changes to how we manage our repository so this won't happen in the future. Apologies for the unnecessary thrashing.)

To build, I had to disable the use of the closure compiler on the release build. I recall from our previous discussion that it isn't essential.

The xscl code is very pragmatic. As you note, it is a combined subset of xsc and xsl - just enough for current purposes. For the long term, we should find a way to eliminate the code duplication as it will invariably lead to maintenance pain. For the short term, your xscl code is a practical path ahead. So, let's keep going. ;)

I'm able to build. I don't have a test harness to be able to run it locally. That seems like the next step, so that I can run and explore. How do you recommend proceeding on that front?

FWeinb:
@phoddie
Sorry for the long delay. I hope i will have some time next week to publish the code for the WebIDE. After that we should collect some requirements and think about ways to use this in the future. 

I think it would be great to have a ready to use image for esp8266/esp32 that one can flash and be good to run code using the WebIDE.

phoddie:
@FWeinb , I'll take a look at getting [`runmod`](https://github.com/phoddie/runmod) working on ESP32. It shouldn't be too difficult.

While unrelated to `wasm`, looking at the WebIDE the most significant missing pieces is debugging. That should be possible, though it will require some changes To be discussed...

phoddie:
@FWeinb, I committed a few small changes to runmod so it works on both ESP8266 and ESP32.

FWeinb:
@phoddie
Great. Will have a look at it. 

Regarding debugging support. If we want to integreat the debugger into the WebIDE it would be great if the debugging protocol could work over WebSockets. I don’t know how much overhead that would mean but that way the WebIDE could connect directly with the running programm and it would be possible to show debugging info/options in the code editor (like defining breakpoints etc)

FWeinb:
@phoddie 
I had some time today to publish my code of the WebIDE to GitHub. You can find it here https://github.com/FWeinb/moddable-webide 

There is some work left to do but it is a starting point.

phoddie:
@FWeinb, thanks for posting the code. It looks straightforward, though I'm not a React or Node programmer, so there is learning to be done. ;)

Regarding debugging, I understand why WebSocket would be a convenient solution here. It is surely possible, but not trivial. Let me share some background...

Our xsbug JavaScript debugger always connects to the debug target over a network socket. When the target device is connected by serial, we run a serial/socket bridge so xsbug only needs to support the network socket interface. So, network debugging is already there. It works on ESP8266 and ESP32, although we more-or-less never use it there.

However... the model is reversed from the browser. The browser wants to be a WebSocket client. The XS network debugging protocol makes the debug target the client and the debugger, xsbug, the server. If the browser could be a WebSocket server, that would simplify things - but I imagine that is not an option, at least not in the general case.

The Moddable SDK supports WebSocket today, both client and server. That is implemented in JavaScript, which won't work for the JavaScript debugging support. Implementing the necessary subset in C is entirely possible, but a small headache because of the handshaking and obfuscation required by the WebSocket protocol to allow it to work reliably on port 80.

As we do for serial, we could run a proxy/bridge server that converts the xsbug native protocol to WebSocket. That would need to run locally or in the cloud. The code isn't complicated. However, it means that the WebIDE is no longer 100% browser hosted, which would is an undesirable outcome.

So... I need a little time to think about options. Maybe there is an elegant solution rather than a brute force one.

FWeinb:
@phoddie Thanks for the in depth explanation. Targeting the serial/network layer in an abstract way like it is implemented know sounds like a logical solution. 

To have this running in the browser today would mean to have some kind of proxy/bridge which would render the browser based approach mood. 

As you said, running a WebSocket server in the browser is not possible. The embedded device would need to handle that and let the browser connect to it. 

I don't now how feasible it would be to implement some kind long polling solution or [Server-Send Events (SSE)](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events). I think SSE would give us a low overhead unidirectional pipe between the device and the browser. Combining this with a (very) primitive HTTP API to interact with the debugging protocol (etc. pause/continue/step etc.) should be enough to make the debugging work. 

Any other ideas?

phoddie:
@FWeinb -- SSE would be perfect, if it worked in both directions. Sockets are expensive enough on ESP8266 that having more than one for debugging is probably not a good idea. The right answer really may be to use a WebSocket with a dedicated implementation. Just need to get comfortable with that idea...

On the good news front, @lprader was able to build (and run) the WebIDE from your sources. Very cool.

FWeinb:
@phoddie Yeah I feared that there were some hardware limitations on using that. Another thing that could be interesting, but would probably need hardware support, is to use [WebUSB](https://wicg.github.io/webusb/) to read the serial data from the device itself. (That would only give use chromium support, as of writing)  

@lprader Great that you could get it up and running! If you have any questions feel free to open an issue or just ask here.

phoddie:
@FWeinb - Yesterday I thought of a way to cleanly bring up the debugging interface to the browser. Now I need to find some time to implement it to see if it works...

WebUSB is spiffy but, as you note, not as generally useful. To support that would, more or less, be a parallel code path for deploy and debugging communication. But, one step at a time. Let's see if we can make this work over Wi-Fi with no wires or browser dependencies first.

phoddie:
@FWeinb - I was able to get the web browser to speak to `runmod` for debugging. There is a proof-of-concept committed now. The proof-of-concept has no user interface at all, just a way to exercise the protocol from script.  I wrote up the details [here](https://github.com/phoddie/runmod#debugging). Establishing the connection is very different from xsbug, but the protocol is identical. Therefore, in theory, the browser should be able to perform all the same debugging operations as xsbug.

FWeinb:
@phoddie
This looks great! Awesome work. Did not think this would be implemented this quickly. I do have a lot of work todo this week so including this in the WebIDE can take me a little bit. 

@lprader are you working on the WebIDE too? If so it might be a good idea to coordinate the development a bit more. 

I think we should move this discussion to an open issue either on the runmod or the WebIDE. What do you think?

phoddie:
@FWeinb - Thanks. It came together faster than I expected too. ;)

Yes, let's move this discussion to issues on the `runmod` repository. This thread is increasingly distant from the original topic.

PR #133, opened 2019-02-25
modules/files for Linux, ESP32

lll000111:
I'm not proficient on these platforms and it's been ages since I programmed C, but all I had to do was pretty much just copy the Mac version, the rest is minor stuff. Very conveniently, ESP-IDF hides SPIFFS (and other FS) internals behind a "vfs" (Virtual filesystem component) and I/O simply uses standard methods such as `stdio`.

For ESP32 I did not forget to add calls to `modInstrumentationAdjust` though, which don't exist in the Mac version.

I ran the ESP32 test on an ESP-WROVER-KIT with 4 MB RAM. I cannot copy any output from the xsbug console (**feature request** - PC platform is Linux, does copy work on Mac or Win?) but here is the result of running examples/files/files/main.js using `mcconfig -d -m -p esp32`:

image

<br>

It also showed the same on Linux (minus the `System.info()` output at the start and the end).

<br>

**Contributor license agreement:** If you find this PR useful and correct, I'll have it filled out and send it to you (the corporate version). Most of it is your own code anyway (Mac version of files-code).


mkellner:
Please submit a CLA (Contributors License Agreement) as describe at the bottom of this page:

https://github.com/Moddable-OpenSource/moddable/tree/public/licenses

Thank you!

lll000111:
<br>

[**Corporate_Contributor_License_Agreement.pdf**](https://github.com/Moddable-OpenSource/moddable/files/2904457/Corporate_Contributor_License_Agreement.pdf)

<br>

lll000111:
**Question**

I just remembered that, to help during development, I had made one "File not found" error message a bit more verbose because I was unable to see anything in xsbug, contents of `path` and other variables:

https://github.com/Moddable-OpenSource/moddable/pull/133/files#diff-febff9205574799087291d955126cacaR79

```c
if (NULL == file)
        xsUnknownError("file not found: %s [write: %u, path: %s]", strerror(errno), write, path);
```

1. For consistency, should this be more like the others, which are all just "file not found"?
2. I really like error codes for certain well-known and frequently expected and used errors such this one. Often this isn't a true exception but expected during program flow, and I don't like checking for a returned message string, they could change (e.g. internationalization), or they could be inconsistent (see right here). As far as I can see I can't send an error code though, only a string?

<br>

**EDIT:**

I added a commit that changes that error message back to the default used everywhere else.

bfriedkin:
Hello -

Thank you much for contributing Files modules for the Linux and ESP32 platforms. We have merged your code into our internal repository and fixed a few small issues along the way. For example, there were a couple of places in the ESP32 static functions where a call to stopSPIFFS() appeared to be missing. But overall things looked good and seem to work well. We also merged your .gitignore change.

I'm glad that you reverted the "file not found" error message back. We actually did the same during our merge. As you noted, it makes sense to be consistent across the implementations, especially since they are all nearly identical, thanks to each platform supporting stdio APIs.

Regarding error reporting, the [Files API](https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/files/files.md) throws  an exception when the file cannot be created and/or opened. Since the exception is thrown by the constructor, your applications could wrap the code that creates a File instance with a try/catch block and act accordingly. The static File.exists() function on the other hand returns a boolean value, as you know, so calling File.exists() in your application might make sense to avoid an exception on open. Also please note that the native code `xsUnknownError` macro causes the current function to exit immediately. It is therefore important to clean up before throwing the exception, e.g. call `stopSPIFFS()` before throwing. We also made these changes in our merge.

If you need a different error reporting mechanism, you could potentially write a custom version of Files that includes such a feature. It is possible to return different types of data from native code XS binding functions. Refer to the [XS in C](https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/xs/XS%20in%20C.md) document for further details.

Thank you again for sharing your contributions! Your code will be available in our next open source update. This pull request will be closed when that happens.

Regards,
Brian

lll000111:
Very nice!

PR #167, opened 2019-04-20
ESP32 - telnet - correct manifest

wilberforce:
remove unused graphics dependancies

### DIFF
--- a/examples/network/telnet/manifest.json
+++ b/examples/network/telnet/manifest.json
@@ -2,9 +2,6 @@
 	"include": [
 		"$(MODDABLE)/examples/manifest_base.json",
 		"$(MODDABLE)/examples/manifest_net.json",
-		"$(MODDABLE)/modules/drivers/ili9341/manifest.json",
-		"$(MODULES)/pins/digital/manifest.json",
-		"$(MODDABLE)/modules/pins/i2c/manifest.json",
 	],
 	"modules": {
 		"*": [
--- a/examples/network/telnet/manifest.json
+++ b/examples/network/telnet/manifest.json
@@ -19,44 +19,4 @@
 		"telnet",
 		"cli",
 	],
-	"platforms":{
-		"esp": {
-			"include": [
-				"$(MODDABLE)/modules/files/file/manifest.json",
-			],
-			"modules": {
-				"commodetto/Bitmap": "$(COMMODETTO)/commodettoBitmap",
-				"pins/digital": "$(MODULES)/pins/digital/digital",
-				"pins/smbus": "$(MODULES)/pins/smbus/smbus",
-				"cli/digital": "$(MODULES)/base/cli/commands/digital",
-				"cli/esp": "$(MODULES)/base/cli/commands/esp/esp",
-				"cli/files": "$(MODULES)/base/cli/commands/files",
-				"cli/i2c": "$(MODULES)/base/cli/commands/i2c",
-				"cli/network": "$(MODULES)/base/cli/commands/network",
-				"cli/pixelsout": "$(MODULES)/base/cli/commands/pixelsout",
-			},
-			"preload": [
-				"file",
-				"pins/*",
-				"cli/*",
-				"commodetto/*",
-			],
-		},
-		"esp32": {
-			"modules": {
-				"commodetto/Bitmap": "$(COMMODETTO)/commodettoBitmap",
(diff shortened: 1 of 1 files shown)


phoddie:
I'm unsure how the manifest is incorrect. It looks like what you've done is removed the ESP8266 and ESP32 specific commands. Would you explain what that fixes.

wilberforce:
It won't build for the ESP32.:

```
C:\Users\rhys\source\repos\telnet>mcconfig -m -d -p esp32
        1 file(s) copied.
# xsc cli\i2c.xsb
# xsc cli\network.xsb
# xsc cli\pixelsout.xsb
# xsc mc\config.xsb
# xsid modTime.c.xsi
# xsid modTimer.c.xsi
# xsid timer.c.xsi
# xsid Resource.c.xsi
# xsid modInstrumentation.c.xsi
# xsid modSocket.c.xsi
# xsid modNet.c.xsi
# xsid modResolve.c.xsi
# xsid modsntp.c.xsi
# xsid xs6wifi.c.xsi
# xsid commodettoBitmap.c.xsi
# xsid digital.c.xsi
# xsid modTimer.h.xsi
# xsid modInstrumentation.h.xsi
# xsid modSocket.h.xsi
# xsid commodettoBitmap.h.xsi
# xsl modules
### ReferenceError: require: module "pins/i2c" not found!
NMAKE : fatal error U1077: 'C:\Users\rhys\Projects\moddable\build\bin\win\release\xsl.EXE' : return code '0xd'
Stop.
```

`ReferenceError: require: module "pins/i2c" not found!`

phoddie:
When I build with this command line...

     mcconfig -d -m -p esp32

...I see errors about the configuration constants for the display driver being missing. For example:

    ~/Projects/moddable/modules/drivers/ili9341/modIli9341.c:95:71: error: 'MODDEF_ILI9341_DC_PIN' undeclared (first use in this function)

That makes sense because the generic ESP32 build target doesn't have a configuration for a display driver.

There are two easy solutions:

1. When building, use a target that configures the display.

```
     mcconfig -d -m -p esp32/moddable_zero
```

2. Remove [the line that includes the ILI9341 driver](https://github.com/Moddable-OpenSource/moddable/blob/6d0ee733046e2de3c0d20991777d94eba6a12bfb/examples/network/telnet/manifest.json#L5) from the manifest, to build with the generic ESP32 target.

wilberforce:
Sorry - I'm missing something here.

Why does the telnet example for the ESP32 reference graphics at all? The only modules used are:

```
import Telnet from "telnet";
import CLI from "cli";
```

phoddie:
> Sorry - I'm missing something here.
>
> Why does the telnet example for the ESP32 reference graphics at all?

The telnet example's manifest references many features available on the ESP32 using the Moddable SDK. Those include files, digital pins (GPIO), I2C bus, and displays. These features are accessed over telnet through corresponding Command Line Interface (CLI) [command modules](https://github.com/Moddable-OpenSource/moddable/tree/public/modules/base/cli/commands) each of which implements a command line interface to a particular feature.

In the specific case of graphics, the [`cli/pixelsout`](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/base/cli/commands/pixelsout.js#L8) CLI command module provides commands to display the dimensions of the display and to fill areas of the screen with color. This is useful when bringing up a display to check the display -- that it displays pixels at all, that the bit order of the colors is as expected, that the orientation is correct, etc.

> The only modules used are:
> 
>     import Telnet from "telnet";
>     import CLI from "cli";

Those are the only modules directly imported by the `main` module. They are not the only modules used. For example, the `cli/digital` command module [installs](https://github.com/Moddable-OpenSource/moddable/blob/6d0ee733046e2de3c0d20991777d94eba6a12bfb/modules/base/cli/commands/digital.js#L25) itself into the CLI module.

wilberforce:
ok - I understand now. Thanks.

I just expected the telnet basic example to compile without modifying the manifest.

phoddie:
> I just expected the telnet basic example to compile without modifying the manifest.

It does when built with a build target that defines a display, as explained above. To use it in an environment that does not define a display does require a change.

phoddie:
FWIW - [this commit](https://github.com/Moddable-OpenSource/moddable/commit/dee8f8f0f3294418901790cf7569f8faf45e4eed) removes the dependency on the display driver in the manifest. If the target provides a display through the `screen` global it is used.

PR #170, opened 2019-04-25
drivers/onewire for Dallas Temp sensors

wilberforce:
This adds onewire support for the ESP32 using the RMT interface. It was been tested with 2 ds18s20 devices on pin 23 on an ESP32. There is also a stub for `-p win` for debugging that should be portable to mac and lin.
There is no documenation at this point - I can add if you want to add to the repository. There are a couple of points that need discussion that might affect the documentation so are waiting until then.


phoddie:
Very nice. Thank you. Will give a review in the next few days.

wilberforce:
Here are  some of the questions I have:

a. Slots usage seems to be a saw tooth - it goes up (even in the win version) and then decreases. I don't have enough experience of the platform to know if this is usual. I don't think there are memory leaks however need to know how to check.

B. `close()` vs destructor. If have copied other examples (like neo pixel) however what if the end user does not call close? Can the de-allocation of resources just occur in the destructor?

C. None of the other examples have comments -

D. The underlying c onewire lib - I had to clone and then flatten the directory structure - ideally I would have though having the lib as a gift hub submodule would be best but don't know how to do this in this environment. ( the .h files are in a include sub folder)
E. Not sure about best way to handle the constructor with regards to the dictionary parameter and then checking those Params.

phoddie:
Let me try to answer your questions now. I'm still making time write up notes on the code.

> a. Slots usage seems to be a saw tooth - it goes up (even in the win version) and then decreases. I don't have enough experience of the platform to know if this is usual. I don't think there are memory leaks however need to know how to check.

This is normal. The garbage collector doesn't run until there's no free space in a heap. You have a problem if it just keeps increasing, or if the GC is running frequently enough to impact performance.

> B. close() vs destructor. If have copied other examples (like neo pixel) however what if the end user does not call close? Can the de-allocation of resources just occur in the destructor?

Because you call `xsRemember` in the constructor, the object will not be collected until either `close` is called which calls `xsForget` making it eligible for garbage collection or the virtual machine is terminated. When the VM ends, the destructor is called on all remaining instances. We typically implement a `close` call on objects which correspond to native resources to have an explicit way to free them as when the garbage collector runs is non-determininistic.

> C. None of the other examples have comments -

The goal has always been to keep the examples small enough that they are more-or-less self explanatory to someone familiar with the material. The documentation should fill in any gaps. If you think it would be useful to include comments in your example, that's fine too.

> D. The underlying c onewire lib - I had to clone and then flatten the directory structure - ideally I would have though having the lib as a gift hub submodule would be best but don't know how to do this in this environment. ( the .h files are in a include sub folder)

We haven't found a way of handling submodules that we like. What you've done matches what we've done in similar situations. For a handful of files, it is a workable. It would be terrible for a big library.

> E. Not sure about best way to handle the constructor with regards to the dictionary parameter and then checking those Params.

At the moment, you appear to just have a `pins` property in the dictionary. The code looks reasonable. What are you unsure of?

wilberforce:
Thanks for the answers a-d. For the constructor - I was more thinking of the DS12x20:
https://github.com/Moddable-OpenSource/moddable/pull/170/commits/951fe29f8fd1cb1a68b2d54e71385dc8e5100c99#diff-6044dd2da22378e439bdb33f7c3d207dR24

connect(new OneWire(pin)) - use the first found DS18X20 device
connect(new OneWire(pin), n) - use the nth DS18X20 device
connect(new OneWire(pin), rom) - use the DS18X20 device with the given rom

A possible future thing might be to add another param to set the desired number of decimal places.

The question was more around type checking and optional params.

barbiani:
@wilberforce How about returning an iterator?

phoddie:
> For the constructor - I was more thinking of the DS12x20

The usual pattern in the Moddable SDK is to have all parameters of the constructor in a dictionary. That often simplifies both calling the constructor and the implementation. Also, it gives labels to the parameters, which improves readability. Borrowing from your example app, if I understand it correctly, you have something like this now:

```
const ow = new OneWire({
  pin: 23
});

const sensorA = new DS18X20(ow, device);  // by ROM ID
const sensorB = new DS18X20(ow, 2);  // by index
```

With a single dictionary argument, the could be explained in the code, rather than in a comment:

```
const sensorA = new DS18X20({wire: ow,, id: device});
const sensorB = new DS18X20({wire: ow, index: 2});
```

The constructor could handle the special cases where there is no `index` or `id` as  `index: 1` and could reject the case where both `id` and `index` are present.

(I'm not familiar with OneWire terminology. Perhaps `bus` (or something else) is a better name for  the `wire` property.)

> A possible future thing might be to add another param to set the desired number of decimal places.

A dictionary makes optional parameters straightforward:

```
const sensorB = new DS18X20({wire: ow, decimalPlaces: 4});
```

jim80:
I was just going to ask a question about DS18X20 support, so thanks!
In a way, that question may have been about adding existing libraries in general, but I think maybe the answers would be beyond me, being more of a js coder than a c one, but I'll check the source of this PR and see if I can get the gist of it. Thanks again!

phoddie:
I started at the bottom of the stack -- with onewire.js and onewire.c. They are pretty reasonable. Most of my notes are details. Let's work through those before proceeding to the higher levels, as changes at the bottom may impact the higher level code.

----

### onewire.js

This won't work:

	skip() {
		xs_onewire_write(0xCC);
	}

I believe it should be:

	skip() {
		this.write(0xCC);
	}


The `rom` function here is behaves as a `toString`:

	// hex version of the rom
	rom(buffer) {
		return BigInt.fromArrayBuffer(buffer).toString(16);
	}

Using `BigInt` here is clever but expensive as it requires the XS engine to include parts of the `BigInt` class support, which could otherwise be excluded to save flash code space (many projects don't don't require `BigInt`). This appears to only be be provided as a convenience for debugging. You might just omit it. Or, you could provide an implementation built on more common JavaScript functions. It should have a name that describes what it does, e.g. `romToString()`. Further, it is not an instance function as it doesn't depend on the class state, so it probably should be a static function. Something like this:

	static romToString(buffer) {
		let result = "";
		(new Uint8Array(buffer)).forEach(i => result += i.toString(16).padStart(2, 0));
		return result;
	}

Similarly, the `crc` function should probably be declared static.

### onewire.c

In `modOneWireRecord` the `the` field is unused.

In `xs_onewire` the call to

	xsmcSetHostData(xsThis, onewire);

Should occur before any exception might be thrown or the memory will be orphaned. This call could throw, if the argument cannot be coerced to an integer, for example:

	pin = xsmcToInteger(xsVar(0));

Similarly, the call to `xsRemember` should only occur after the constructor has succeeded, since once it is called the object cannot be garbage collected until xsForget is called, which will never happen if the constructor exits by throwing.

The call to `owb_rmt_initialize` doesn't check its result. It looks like that has the possibility to fail. In my experience working with hardware drivers, initialization calls are the most likely to fail so it is good to check for that here. 

In a few places the code does this:

	int argc = xsmcToInteger(xsArgc);

That works, but it is more efficient to write:

	int argc = xsmcArgc;

In `xs_onewire_write` the code pins invalid values. That's reasonable, but it might be better to fail on invalid values to warn the caller that something is wrong:

	uint8_t value = xsmcToInteger(xsArg(0));
	if ((value < 0) || (value > 255))
		xsRangeError("bad value");

The design of the `search` function is to return all values in a single array. That is no problem if the number of devices is reasonably small. As the number increases, the amount of memory required increases in which case it may be preferable to return a single result at a time. As Marcio Barbiani suggested, an iterator would well for that, though the implementation is a bit trickier. It may be worth the effort at this point. If the implementation stays as is, there is no need to create a temporary variable on the XS stack, as you can just use xsResult directly:

	xsResult = xsNewArray(0);

	while (found)
	{
		xsCall1(xsResult, xsID_push, xsArrayBuffer(&search_state.rom_code.bytes,8));
		owb_search_next(onewire->owb, &search_state, &found);
	}

The `xs_onewire_reset` code returns an integer to the script though it appears that the native `owb_reset` returns a boolean value.

	xsmcSetBoolean(xsResult, present);

wilberforce:
Thanks for the detailed review.

I've done most of what you suggested:
https://github.com/Moddable-OpenSource/moddable/pull/170/commits/456309e4ae71e5a738ca0a30df52ba2160fd61e0


a.  _Static functions_
I could make the `Onewire.hex` static but struggled with `Onewire.cfc`.
In the DS18X20.js  class If I use:
`if (OneWire.crc(s.buffer, 8) == s[8]) {`
I get:
`xs_onewire_select: xs_onewire_select: C:\Users\rhys\Projects\moddable\modules\drivers\onewire\DS18X20.js (108) # Break: (host): C: xsGetHostData: no host object!`
This seems to be scope related?

b.  _Hex version of buffer_
> This appears to only be be provided as a convenience for debugging. You might just omit it. 

You might need save as a  text version of device  id in configs:
```
probes: {
   inside_fridge: '28CC2E230500006B',
   outside_fridge: '283260DC04000001'
}
```

c. _owb_rmt_initialize doesn't check its result._
It doesn't look like the base lib returns null on fail. I think the main reason it would fail is if the RMT channels were already in use. I have purposely advance on from channel 0 that neopixels use to avoid the confict and this can be controlled by defined in the manifest.

So in this case is it worth moving `xsRemember(onewire->obj);` ?

d. _search function is to return all values in a single array._
The bus supports up to 127 devices on a single pin.  Typically it is 1-3 devices,  certainly under 10.

This could be updated later to an interator if required. It think it would be fine for now.

wilberforce:
update the ds18x20 to use the dictionary constructor.

Still have the issue of the static OneWire.hex function

https://github.com/Moddable-OpenSource/moddable/pull/170/commits/e0f3d8acdfaab93d4d8686ba5bf6ffe839227939

phoddie:
Looking better. Thanks for explaining the use of saving ROM ID strings. 

Regarding the problem with `OneWire.crc`, it looks like it isn't declared static in [OneWire.js](https://github.com/Moddable-OpenSource/moddable/blob/456309e4ae71e5a738ca0a30df52ba2160fd61e0/modules/drivers/onewire/onewire.js#L43):

```
crc(buffer) @ "xs_onewire_crc";
```

instead of:

```
static crc(buffer) @ "xs_onewire_crc";
```

Regarding `OneWire.crc`, once it is a static function it isn't part of the instance. That means it doesn't have the host data of an instance. In fact, as declared, it has no host data. But, that's fine. You don't need it. Your ESP32 version doesn't call `xsmcGetHostData` and while your Windows version does, it doesn't use the result.

The placement of `xsmcSetHostData` and `xsRemember` isn't right yet. More or less, `xsmcSetHostData` should go as early as possible and `xsRemember` as late as possible. Something like this:

```
void xs_onewire(xsMachine *the)
{
  modOneWire onewire;

  xsmcVars(1);

  if (!xsmcHas(xsArg(0), xsID_pin))
    xsUnknownError("pin missing");

  onewire = c_malloc(sizeof(modOneWireRecord));
  if (!onewire)
    xsUnknownError("no memory");

  xsmcSetHostData(xsThis, onewire);

  onewire->obj = xsThis;

  xsmcGet(xsVar(0), xsArg(0), xsID_pin);
  onewire->pin = xsmcToInteger(xsVar(0));

  // Create a 1-Wire bus, using the RMT timeslot driver
  onewire->owb = owb_rmt_initialize(&onewire->rmt_driver_info, onewire->pin, MODDEF_ONEWIRE_RMT_TX_CHANNEL, MODDEF_ONEWIRE_RMT_RX_CHANNEL);
  owb_use_crc(onewire->owb, true); // enable CRC check for ROM code

  xsRemember(onewire->obj);
}
```

phoddie:
I'm starting to read the `DS18X20` class in detail. The `searchAlarm` function doesn't seem right. First, it is a method on a DS18X20 instance but it returns information about active alarms for multiple devices. That suggests it should be a function of a `OneWire` instance like `search`.

The implementation of `searchAlarm` calls `search` with an argument of 0xEC which is unused by its implementation in `xs_onewire_search`.

```
searchAlarm() {
 return this.bus.search(0xEC);
}
```

The value 0xEC seems to be from `omb.h` just as the 0xCC constant in `skip`:

```
#define OWB_ROM_SEARCH        0xF0
#define OWB_ROM_READ          0x33
#define OWB_ROM_MATCH         0x55
#define OWB_ROM_SKIP          0xCC
#define OWB_ROM_SEARCH_ALARM  0xEC
```

Maybe then xs_onewire_search should be something like:

```
void xs_onewire_search(xsMachine *the)
{
  modOneWire onewire = xsmcGetHostData(xsThis);

  uint8_t value = OWB_ROM_SEARCH;
  if (xsmcArgc)
    value = xsmcToInteger(xsArg(0));
  owb_write_byte(onewire->owb, value);

// same as before
}
```

And `searchAlarm` moves to `onewire.js` as

```
searchAlarm() {
 return this.search(0xEC);
}
```

wilberforce:
@phoddie 

The issue is the alarm search does not make sense in the onewire.js - as it relates to temperature sensors. If you where using a onewire  eeprom - there would not be any devices raising alarms.

I added the alarm search for completeness.,. some libraries provide it - however given we have a javascript api and can handle alarms in the application, It may be better off to just to dump this functionality? This then avoids the issue. Do you agree?

wilberforce:
Looking at this beauty of design:

http://owfs.sourceforge.net/simple_commands.html

It seems there are other devices that do need the search - so perhaps your suggestion is the best.

wilberforce:
Another commit...
a. removed the alarms - as the base code did not allow changing the search command without modifying the code and for the reasons above decided to keep it simple.
b. The win simulator returns fake temps
c. Made ` static OnWire.crc()`

wilberforce:
Added initial ESP8266 port using ModGPIO drivers. Unfortunately it crashes:

```
Exception (29):
epc1=0x402058f8 epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000
ctx: cont
sp: 3ffedcd0 end: 3ffee3b0 offset: 01a0
>>>stack>>>
3ffede70:  3fff84bc 3fff8e84 3fff88f4 40204f2e
3ffede80:  3fff849c ffffffff 3fff88f4 00000001
3ffede90:  ffffffff 3fff84bc 4025caab 00000001
3ffedea0:  ffffffff 3fff84bc 4025caab 4021859c
3ffedeb0:  00000000 000000ff 4025c8f0 00000000
3ffedec0:  3fff88f4 3fff84fc 3fff85fc 40264828
```

```
root@office-pc:/mnt/c/Users/rhys/Projects/moddable/build/tmp/esp/debug/onewire# addr2line -e main.elf 0x402058f8
/cygdrive/c/Users/rhys/Projects/moddable/examples/drivers/onewire/C:\Users\rhys\Projects\moddable\modules\drivers\onewire\esp/owb_gpio.c:128
root@office-pc:/mnt/c/Users/rhys/Projects/moddable/build/tmp/esp/debug/onewire#
```

https://github.com/wilberforce/moddable/commit/fd3fa335054de16e8fd8baca9f2922aa9cbdb53a#diff-dd09fb1a0ea0859425c9734a81b03ad4R128
128: `    *is_present = present;`

So like like stack is corrupting.

phoddie:
Naive question: assuming the line number is correct, is it possible `is_present` is NULL?

wilberforce:
> Naive question: assuming the line number is correct, is it possible is_present is NULL?

Not Naive at all!

I'll check.

Also realised I had not called ` modGPIOInit` so that might the issue... Need to look at this code:
https://github.com/Moddable-OpenSource/moddable/blob/26bb48a3857976b2be9a1087376914d4fa79c705/modules/pins/digital/digital.c#L121

phoddie:
My DS18X20 sensor arrived a couple days ago. I wired it too an ESP32 tonight and finally got a chance to run your code. It works nicely! Once it was running, it was much easier to tinker with the code. A few notes:

- The `present` property on the sensor seems more like a OneWire bus function. I implemented `OneWire.prototype.isPresent(id)`. It is called from `DS18X20.prrototype.get present` though that's a convenience and maybe should be eliminated. The native implementation of `isPresent` is lighter than using `search` as it avoids allocating any memory and can terminate the search when it finds the target id.
- The `digits` property of the constructor seems out of place. When used, it changes the returned temperature value from a number to a string because `toFixed` returns a string. That seems potentially dangerous. I eliminated all that and put the call to `toFixed` in the example when it outputs the temperature.
- The `DS18X20` module wasn't preloaded.
- I moved `max_conversion_ms` to the global scope of the module. That puts it in flash memory which avoids allocating it on each callback based temperature reading.
- Some properties go by multiple names (rom/id, s/scratchpad, bus/ow). I made these consistent which makes it easier to read (for me, at least!) and allows for concise object declarations.
- The `temp` getter, `read` getter, and `getTemp` function had me confused for a while. Once I understood it, I appreciate the functionality it provides - getting the most last reading or getting the next reading as soon as possible. `getTemp` and the `temp` getter overlap to the point where there should only be one. Since a getter cannot have a parameter, I suggest `getTemperature()` with an optional callback.

I'll sent you the changes out-of-band (email). If you have any questions, just let me know. I'm happy to try to explain my thinking.

wilberforce:
I (hopefully) have added your changes to the pull request - please take a look here:
https://github.com/Moddable-OpenSource/moddable/commit/aae96e02b7427a3be74092996ee68d0d76723904

- `present`. I also wonder if the property needs to on the DS18X20. Shall we ditch it?

- `read` is a private method. I have renamed as `_read`, however it could be a function within `getTemperature()`. I like the idea of the property `.temperature` and dislike function names with `get`, however there is no choice for the callback... 
The argument to the callback is only the actual temp, - I wondering if it should be the instance - so it also had the id with it? 

- ` max_conversion_ms[this.resolution]); //@@jph this always retreives the resolution. necessary? Or can that be cached?`
If the callback is not used, then this is probably not going to get called. If it gets cached in `this._resolution` it would need to be set in the constructor and if the setter was called. I'm not sure it's worth the trouble?

- We haven't discussed the alarms mentioned above. They would be set in the DS18x20 but the search would need to be modifed in Onewire. I'm not sure if this adds value?

wilberforce:
I suppose this could be done:

`max_conversion_ms[ _this.resolution | this.resolution]`

```
get resolution() {
   this._resolution=[0x1F, 0x3F, 0x5F, 0x7F].indexOf(this.scratchpad[4]) + 9;
    return this._resolution;;
  }
```

and in the setter to force a new read.

`this._resolution=null;`

wilberforce:
The esp8266 port is working... it read temps but does crash with a stack trace:

```
Exception (0):
epc1=0x40000527 epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000079 depc=0x00000000
ctx: sys
sp: 3ffffbe0 end: 3fffffb0 offset: 01a0
>>>stack>>>
3ffffd80:  3ffffd90 00000000 0000043e 00000000
3ffffd90:  40000f68 00000030 0000001e ffffffff
3ffffda0:  40000f58 00000000 00000020 00000000
3ffffdb0:  3ffe9990 c063075c 3ff20a00 c0000000
3ffffdc0:  00003000 00000000 00000058 3fffdab0
3ffffdd0:  00000000 3fffdcc0 3ffe9320 00000030
3ffffde0:  00000000 00000000 00000020 4010316e
3ffffdf0:  3ffe9280 4023472b 3fff072c 40201dc4
3ffffe00:  00000000 4023fdb0 3ffeebf8 40105ecc
3ffffe10:  0214b439 3ffef7e4 00000000 402516f6
3ffffe20:  00000000 00000000 401056a1 00000100
3ffffe30:  00000058 00000001 40102926 3ffe8e40
3ffffe40:  3fff072c 40102d06 00000001 00000000
3ffffe50:  4022d918 40102ad7 00000000 00000000
3ffffe60:  40102b12 00000000 00000002 3fffdcb0
3ffffe70:  402343ee 3fffdab0 00000000 40201053
3ffffe80:  3ffe9320 40000f49 3fffdab0 40000f49
3ffffe90:  40000e19 000000d3 00000002 3ffffed0
3ffffea0:  3ffffed8 ffffffff ffffffff 40100270
3ffffeb0:  40100276 00000002 3ffffed0 3ffffed0
3ffffec0:  4010000d 00000004 00000008 00000000
3ffffed0:  40106740 000018c0 00000000 d3000000
3ffffee0:  00000000 400042db 40106738 00000001
3ffffef0:  00000008 40100004 00000001 401000b5
3fffff00:  00001010 003ff000 3fffffb0 00000020
3fffff10:  00000673 00000000 00073680 3ffe8000
3fffff20:  00006740 5ffffe00 00000001 401000a0
3fffff30:  5ffffe00 5ffffe00 000003fc 401009d7
3fffff40:  02040005 02050505 05040005 05050405
3fffff50:  fffdfe04 e0f0f0f0 0ae1e0e0 00f8ffff
3fffff60:  4e52f8f8 3840444a 01010000 05040302
3fffff70:  00000001 00020000 00000000 00000000
3fffff80:  00000ae1 00000000 93010000 00000043
3fffff90:  00000000 00000000 00000000 00000000
3fffffa0:  00000000 00000000 21000000 00000000
<<<stack<<<
```

latest changes:
https://github.com/Moddable-OpenSource/moddable/commit/99f934d4456469e969831f6dadf53690acfab1e2

Here is the trace from the digital scope:

image

image


I'm wondering if the watchdog timer is kicking in here because of the `modCriticalSectionBegin()`, `modCriticalSectionEnd()` taking too long?

phoddie:
> The esp8266 port is working

Excellent. Congrats!

> I'm wondering if the watchdog timer is kicking in here...

Possible, but usually you see an indication of that in the crash log. You can feed the watchdog timer in your loop to see if that eliminates the crash.

```
modWatchDogReset();
```

Can you run `addr2line` against the stack trace to get a hint about where it is dying?

(Will make time later this week to catch up on the discussion above. A bit busy at the moment. Also ordered another temperature sensor to be able to try more than one at at time.)

wilberforce:
The `addr2line` syntax seems different from the esp32 -  however I did get this which was enought of a clue:

```
root@office-pc:/mnt/c/Users/rhys/Projects/moddable/build/tmp/esp/debug/onewire# addr2line -e main.elf 0x402303d3 3ffee090:  3fff5abc 3fff803c 40262ebd 4021d1ac
/cygdrive/c/Users/rhys/Projects/moddable/examples/drivers/onewire/C:\Users\rhys\Projects\moddable\build\devices\esp\lib\tinyprintf/tinyprintf.c:500
```

It was crash in a diagnostic rom sprintf ( "%0x") thing that was not even getting used... so it not crashing any more...

wilberforce:
Last night I compiled the ESP32 vesion of the library with the `ModGpio` version used by the esp8266 version to check it out. ( This ESP9266 version is crashing with a stack trace - this seems to be network related rather than to this lib).
I added to the `adafruit.io` example and logged temperatures every 60 seconds through the night:

image

It works well, with a small issue to resolve.

time|temp C|
------------ | -------------
2019/05/10 12:54:58am | 23 | 
2019/05/10 12:53:58am | 23 |  
019/05/10 12:52:58am | 23 | 
2019/05/10 12:51:58am | 85 |
2019/05/10 12:49:58am | 23 |
2019/05/10 12:48:58am | 23 |
2019/05/10 12:47:58am | 23.0625 |   
2019/05/10 12:46:58am | 23

1. 85c is when the temperate registers are all 1's. I would imagine the registers either side are 1 as well so might be able to detect this condition and invalidate the crc so the value gets ignored. This sometimes happens on chip power up too for the first reading. 2019/05/10 12:50:58am is missing so perhaps the crc failed on this reading...

2. I'm having trouble with the ESP8266 build in general. I'm getting a crash out even when removing the onewire lib ( which seems stable for the gpio on ESP32). Hello world works ok. Anything network related seems to crash - https://github.com/Moddable-OpenSource/moddable/pull/184

phoddie:
What's the easiest way to reproduce a crash you are seeing on ESP8266? I can find a few minutes to take a look. (FWIW - I've done lots of networking recently on ESP8266 and it has been stable. But, you may well be doing something different that triggers a problem)

wilberforce:
@phoddie
I believe this is ready to commit. The esp8266 crash was not related to this, it was related to switch between 1mb and 4mb variants.

phoddie:
Agreed.

wilberforce:
Commit https://github.com/Moddable-OpenSource/moddable/commit/878314f74b0631054144bf46ce3ce0edb3bf75e9

wilberforce:
@mkellner 

Sorry - I missed two files please see 0e0f475

wilberforce:
@phoddie

Please can we add the two files in the comment above - they are the rmt versions of the driver for esp32

PR #184, opened 2019-05-09
fix double config: streaming-adafruit-io

wilberforce:
The double config overwrites the adafruit io settings

### DIFF
--- a/examples/network/services/streaming-adafruit-io/manifest.json
+++ b/examples/network/services/streaming-adafruit-io/manifest.json
@@ -8,6 +8,7 @@
 		"username": "YOUR_USERNAME_HERE",
 		"feedKey": "YOUR_FEED_KEY_HERE",
 		"AIOKey": "YOUR_AIO_KEY_HERE",
+		"sntp": "pool.ntp.org",		
 	},
 	"modules": {
 		"*": [
@@ -22,8 +23,5 @@
 		"*": [
 			"$(MODULES)/crypt/data/ca170",
 		]
-	},
-	"config": {
-		"sntp": "pool.ntp.org",
 	},
 }


wilberforce:
Should this work in the esp8266?

It's crashing:

```
<<<stack<<<
Exception (9):
epc1=0x402498ee epc2=0x00000000 epc3=0x00000000 excvaddr=0x0000000f depc=0x00000000
ctx: sys
sp: 3ffffc90 end: 3fffffb0 offset: 01a0
>>>stack>>>
3ffffe30:  3ffeba30 00000078 00000000 00000038
3ffffe40:  40242c15 3fff048c 40249c77 3ffecc90
```

wilberforce:
@phoddie > What's the easiest way to reproduce a crash you are seeing on ESP8266? I can find a few minutes to take a look. (FWIW - I've done lots of networking recently on ESP8266 and it has been stable. But, you may well be doing something different that triggers a problem)

This example is crashing for me on the esp8266.

https://github.com/Moddable-OpenSource/moddable/tree/public/examples/network/services/streaming-adafruit-io

The sntp add to config that was done here - https://github.com/Moddable-OpenSource/moddable/commit/3c1a74824d7b469460ad41631c619a0d91efe567

The fix above needs to done to the manifest.

wilberforce:
Got to the bottom of the crashing.

I had put a 1mb image on the board when testing the sonoff. This changed the flash map to 512/512.

Putting the flash map back to `4MB-c1` 1024/1024 sorted the crashing.

wilberforce:
@phoddie 

Are you able to merge this pull request - it fixes the double config in the manifest.

Thanks!

phoddie:
Yes, we'll do that. Thanks for the patch and the reminder.

wilberforce:
Commit https://github.com/Moddable-OpenSource/moddable/commit/c3fa64aea4028cad8245ce875b03277a62a12c93

PR #187, opened 2019-05-13
Add driver for HD44780 character LCD with i2c interface

wilberforce:
This has been tested on ESP8266 - should work with pin change on ESP32


wilberforce:
Arggh. contains One wire changes. I'll need to do a clean pull request...

PR #188, opened 2019-05-13
Add driver for HD44780 character LCD with i2c interface

wilberforce:
This adds the Hitachi character LCD module. It will work with 2x16 and 4x20 displays

IMG_5986


wilberforce:
I''m wondering if these should be changed:

```
var LCD = {};

LCD.CLEARDISPLAY = 0x01;
LCD.RETURNHOME = 0x02;
```

to

```
const LCD_CLEARDISPLAY = 0x01;
const LCD_RETURNHOME = 0x02;
```

There will need to be some object freezing too.

phoddie:
This looks good. The implementation is consistent with many of the patterns of other drivers in the SDK.

The `LCD.CLEARDISPLAY` style constants are reasonable. We do that in other places.

We'll merge this PR as part of our next update.

It might be worthwhile to include the wiring used between the HD44780 display and the PCF8574P expander as a comment in `main.js`?

Issue #192, opened 2019-05-15
I2C issue with M5Stack STEPPER MOTOR module


zBuffer:
I've tried to communicate with the [STEPPER MOTOR](https://docs.m5stack.com/#/en/module/stepmotor) module using the following code, it kept getting `0xFF` response from the device. I've tried changing the bus speed slower/faster, but no avail. For sending, it doesn't work as well, with exceptions thrown (except for `i2c.write()` without parameters).

I've compared the MicroPython [driver](https://github.com/m5stack/M5Stack_MicroPython/blob/master/MicroPython_BUILD/components/micropython/esp32/machine_hw_i2c.c) and [this](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/pins/i2c/esp32/modI2C.c), they don't look that different to me.

Then I looked at their GRBL [firmware](https://github.com/m5stack/stepmotor_module/blob/master/Firmware%20for%20stepmotor%20module/GRBL-Arduino-Library/serial.cpp) and [schematics](https://docs.m5stack.com/assets/img/product_pics/module/stepmotor_sch.png) - which the only curiosity I could spot is the GPIO16, GPIO17 connection, possible for the alternative UART firmware.

```javascript
import I2C from "pins/i2c";
import Poco from "commodetto/Poco";
import Resource from "Resource";
import Timer from "timer";

const render = new Poco(screen);
const font = new Resource("8x8font.dat");
const i2c = new I2C({ address: 0x70 });
const black = render.makeColor(0, 0, 0);
const white = render.makeColor(255, 255, 255);

let lines = [""];
let renderedLines = 0;
let blink = true;
let question = false;

render.begin();
render.fillRectangle(black, 0, 0, render.width, render.height);
render.end();

Timer.repeat(() => {
    while (1) {
        let arr = i2c.read(1);
        if (arr === undefined) {
            question = false;
            break;
        }
        if (arr[0] == 0xFF) {
            question = true;
            blink = !blink;
            break;
        } else if (arr[0] == 0x00) {
            question = false;
            blink = !blink;
            break;
        } else {
            lines[lines.length - 1] += String.fromCharCode(arr[0]);
            if (String.fromCharCode(arr[0]) == "\n") {
                lines.push("");
                break;
            }
        }
    }


    for (let i = renderedLines, y = renderedLines * 9; i < lines.length; i++ , y += 9) {
        render.begin(0, y, render.width, 8);
[...]

PR #194, opened 2019-05-17
Disable bypass in MPU6050 driver and expose I2C Master Bus

zBuffer:
Enabling bypass in MPU6050 causes problems for non-SMBus devices sitting on the I2C bus, as seen in #192 . The slave devices connected to MPU6050 can be accessed indirectly using `MPU6050.I2CMasterBus`, which the prototype is similar to `SMBHold`.

### DIFF
--- a/modules/drivers/mpu6050/mpu6050.js
+++ b/modules/drivers/mpu6050/mpu6050.js
@@ -152,7 +152,7 @@ class Gyro_Accelerometer extends SMBHold {
                 this.config = dictionary;
             }
 
-            calcLength(...value) {
+            calcLength(value) {
                 let len = 0;
                 for(let i in value) {
                     switch(typeof(i)) {
--- a/modules/drivers/mpu6050/mpu6050.js
+++ b/modules/drivers/mpu6050/mpu6050.js
@@ -27,6 +27,10 @@ import SMBus from "pins/smbus";
 import Timer from "timer";
 
 const REGISTERS = {
+    I2C_MST_CNTRL: 0x24,
+    I2C_SLV0_ADDR: 0x25,
+    I2C_SLV0_REG: 0x26,
+    I2C_SLV0_CTRL: 0x27,
     INT_BYPASS: 0x37,
     ACCEL_XOUT: 0x3B, //big endian
     ACCEL_YOUT: 0x3D,
@@ -35,6 +39,9 @@ const REGISTERS = {
     GYRO_XOUT: 0x43,
     GYRO_YOUT: 0x45,
     GYRO_ZOUT: 0x47,
+    EXT_SENS_DATA_00: 0x49,
+    I2C_SLV0_DO: 0x63,
+    USER_CNTRL: 0x6A,
     PWR_MGMT_1: 0x6B,
     PWR_MGMT_2: 0x6C,
     WHO_AM_I: 0x75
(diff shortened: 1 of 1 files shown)


phoddie:
The implementation is using more memory than needed. Since it targets microcontrollers, I would like to look at rearranging the code a bit to reduce that. The `I2CMasterBus` getter creates the class dynamically, which requires RAM for each method of the class, plus execution time at runtime to build the complete class. Instead, I think you can declare the class at the top level as usual:

```
class I2CMasterBus {
	constructor(dictionary, i2c) {
	 this.config = dictionary;
  	 this.i2c = i2c;
	}

	calcLength(value) {
	 let len = 0;
	 for(let i in value) {
		 switch(typeof(value[i])) {
			 case "number":
				 len++;
				 break;
			 case "string":
				 len += value[i].length;
				 break;
			 default:
				 if(value[i] instanceof(ArrayBuffer))
					 len += value[i].byteLength;
				 else if(value[i].length !== undefined)
					 len += value[i].length;
				 else
					 throw "MPU I2CMasterBus unsupported type";
		 }
	 }

	  return len;
	}

	readBlock(register, count, buffer) {
	 if(count > 24)
		 throw "MPU I2CMasterBus readBlock maximum length is 24 bytes";
	 this.i2c.writeByte(REGISTERS.I2C_SLV0_ADDR, this.config.address | 0x80);
	 this.i2c.writeByte(REGISTERS.I2C_SLV0_REG, register);
	 this.i2c.writeByte(REGISTERS.I2C_SLV0_CTRL, I2C_SLV0_EN | count);
	 Timer.delay(1);
	 return this.i2c.readBlock(REGISTERS.EXT_SENS_DATA_00, count, buffer);
	}

	writeBlock(register, ...value) {
	 let count = this.calcLength(value);
	 if(count > 4)
		 throw "MPU I2CMasterBus writeBlock maximum length is 4 bytes";
	 this.i2c.writeByte(REGISTERS.I2C_SLV0_ADDR, this.config.address);
	 this.i2c.writeByte(REGISTERS.I2C_SLV0_REG, register);
	 this.i2c.writeBlock(REGISTERS.I2C_SLV0_DO, ...value);
	 return this.i2c.writeByte(REGISTERS.I2C_SLV0_CTRL, I2C_SLV0_EN | count);
	}

	writeByte(register, value) {
	 return this.writeBlock(register, value & 0xFF);
	}

	writeWord(register, value, endian) {
	 if (endian)
		 return this.write(register, (value >> 8) & 255, value & 255);
	 else
		 return this.write(register, value & 255, (value >> 8) & 255);
	}

	readByte(register) {
	 return this.readBlock(register, 1)[0];
	}

	readWord(register, endian) {
	 let value = this.readBlock(register, 2);
	 return endian ? (value[1] | (value[0] << 8)) : (value[0] | (value[1] << 8));
	}
};
```
Notice that the `i2c` instance is passed to the constructor and stored in `this.i2c`, so all occurrences of `that` are replaced with `this.i2c`.

Then the getter just needs to connect the i2c instanced to the class constructor returned, something like this:

```
    get I2CMasterBus() {
         const i2c = this;
         return class extends I2CMasterBus {
             constructor(dictionary) {
                 super(dictionary, i2c);
             }
      }
```

I don't have the hardware to run this myself, so I can't try it out. If you can, please let me know how it goes. Thanks.

andycarle:
Hi @zBuffer,

I'm unable to test this change, because the revision of the M5Stack Fire hardware that I have is the one with the MAG3110 sensor connected directly to the ESP32's i2c bus. (See this note in the [M5Stack schematics repo](https://github.com/m5stack/M5-Schematic/blob/master/Core/hardware_difference_between_cores.md):  `Be careful, the early version of M5Fire was built in MPU6050 + MAG3110, but we change MPU6050 + MAG3110 to MPU9250.`)

I had initially configured the MPU6050 driver to use the bypass mode so that the various revisions of the M5Stack Fire (and Gray, etc.) hardware could use the same MPU6050 and MAG3110 drivers as each other. But, as you note, this seems to not work with all of the M5Stack accessories.

Given the wide array of M5Stack configurations and revisions (and the variety of projects developers are building with them), I think it would be best to organize this change so that it is optional for users of the mpu6050 module and so that we can easily turn the i2c bypass on and off for the different development targets. 

I suggest that by default we leave the configuration as it was, with the bypass enabled and the internal i2c master turned off. Then modify the `configure(dictionary)` function in the `Gyro_Accelerometer` class to optionally turn off the bypass and enable the internal i2c bus functionality.

Cheers,
&nbsp;\- Andy

PR #204, opened 2019-05-28
Documentation: note that shell needs to be restarted after editing ~/.bashrc

jernst:


jim80:
or maybe "source ~/.bashrc"  ?

https://stackoverflow.com/questions/2518127/how-do-i-reload-bashrc-without-logging-out-and-back-in.

But which ever way, it's not a bad point in a getting started guide - even after 30 years I'll still occasionally forget to restart/refresh the shell by some means when I need to!

Alhadis:
Restarting the shell shouldn't be necessary if you're using `export PATH` after modifying it:

~~~console
λ type foo
-bash: type: foo: not found
λ mkdir -p /tmp/path/to/executable
λ printf 'echo "It works."\n' > $_/foo
λ chmod +x /tmp/path/to/executable/foo
λ export PATH=/tmp/path/to/executable:$PATH
λ type foo
foo is /tmp/path/to/executable/foo
λ foo
It works.
~~~

It's different if you export the `PATH` *after* a program that needs it has finished running, but that's a given.

PR #241, opened 2019-08-03
Change xsTrace ->  xsUnknownError

wilberforce:
Update so that write on a closed socket can be caught on ESP32

### DIFF
--- a/modules/network/socket/lwip/modSocket.c
+++ b/modules/network/socket/lwip/modSocket.c
@@ -619,7 +619,7 @@ void xs_socket_write(xsMachine *the)
 			xsResult = xsInteger(0);
 			return;
 		}
-		xsTrace("write on closed socket\n");
+		xsUnknownError("write on closed socket\n");
 		return;
 	}
 


phoddie:
This code originally threw an exception. That was changed to a trace because the exceptions were causing more problems than silently ignoring the write. Delivery to scripts of data received messages are prioritized over socket closed messages so that all received data gets to the script before the close (receiving data after the close seems very unexpected). Sometimes in the function that is invoked when data is received performs a write. If it does, that will cause an exception with the change you propose because the script does not know (yet) that the socket is closed. (Had it called `write` with no arguments it would have known that 0 bytes can currently be output and waited.)

The trace is a warning of a non-critical error. If there 's a strong reason to change it, I'm open to that. But I'm unsure that there's a single answer that's always right, and experience tells me to prefer the trace.

wilberforce:
Thanks. This is the situation:

- Websocket server running one ESP32
- Browser Client connects to websocket server
- messages are sent bi-directionally - all good
- Browser is closed
- ESP32 server does not know socket is closed and sends data
- The close method on the socket is not called, so ESP32 keeps on trying to send data to a client that is no longer listening.

This is how I'm handling the send:

```
/ if me undefined - send to all
	static broadcast(msg, me) {
		clients.forEach(client => {
			try {
				if (client != me) client.write(msg);
			} catch (error) {
				ModelServer.removeClient(client);
			}
		});
	}
```

All other implementation other than ESP32 use `xsUnknownError` which throws the exception.

With `XsTrace` the error does not bubble up, and the socket close message is not trigger so the code never knows the client has disconnected.

Does error handling need to be added here?
https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/websocket/websocket.js#L73-L96

If the `write` fails - call `disconnect` ?

phoddie:
> ESP32 server does not know socket is closed and sends data

Why doesn't the ESP32 server know? The browser should send a WebSocket close message, which is [handled](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/websocket/websocket.js#L247) by the WebSocket implementation. If it doesn't, the socket itself is closed. That is also [handled](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/websocket/websocket.js#L269) in the WebSocket implementation.

wilberforce:
On the ESP32, After establishing a websocket connection, if I quit the chrome browser from the desktop, the events occur as you describe above. However if I connect from safari or chrome on the iphone, and close the browser window, the socket close events don't get triggered, the first time the server knows about it is when it tries to do the next `write`.

in xsbug:
`....\Projects\moddable\modules\network\socket\lwip\modSocket.c (623) # Exception: (host): write on closed socket`
If this is just an XsTrace, then it will keep on writing to the closed socket everytime the server attempts to send. 

When using the windows simulator - and closing the browser on the phone, this code is triggered:

```
				case 8:
					this.state = 4;
					this.callback(Client.disconnect);		// close
					this.close();
					return;
```

So it seems related to the lwip code on the ESP32.

phoddie:
That you receive the `write on closed socket` trace means that the native socket knows the socket is gone. That happened either because of a clean TCP disconnect (which passes through line 1029 in `didReceieve`) or a TCP error (which passes through line 988 of `didError`).

> ...if I quit the chrome browser from the desktop, the events occur as you describe above...
> However if I connect from safari or chrome on the iphone, and close the browser window, the socket close events don't get triggered

I'm guessing that Chrome on the desktop does a clean TCP disconnect whereas on iPhone the connections are being aborted.

The WebSocket implementation only looks for disconnect messages and ignores error messages. That's a mistake. It should treat errors as a disconnect. 

Would you try a fix? At line [268](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/websocket/websocket.js#L268) and line [405](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/websocket/websocket.js#L405), change

```
     if (-1 == message) {
```

to

```
     if (message < 0) {
```

With a little luck, that will solve the problem.

wilberforce:
@phoddie 

Thanks - I'll give that a try.

wilberforce:
https://github.com/Moddable-OpenSource/moddable/commit/95bce67a1ba65869c30d2523f5378a13dd516038

PR #245, opened 2019-08-11
M5 Stick C - add Gyro_Accelerometer sh200q

wilberforce:
Modify M5 Stack demo.  With the small screen its a bit odd.

- [ ] Need to add dictionary to set acc odr, gyro odf and gyro dlpf

- [ ] GYRO_SCALER  and ACCEL_SCALER need to be checked

- [ ] reading of temperature value - need to bring into global scope like `power` for the AXP192 ?


wilberforce:
Not really sure how the read of the temperature should be handled.

This is reading correctly when uncommented in setup-target.js:

`trace( 'The Temp:',state.accelerometerGyro.sampleTemp(),'\n');`

`The Temp:21`

The temperature value is no going to change as quickly as x,y,z so a callback is probably not required.

andycarle:
@wilberforce I've cherry-picked these two commits over to the public tree then made some edits to the driver and example. They will be in our next open source push. Once those changes are live, please give them a try and let me know if everything is working as expected.

wilberforce:
thanks @andycarle 

I'll see if I get a chance this weekend - as the ESP sdk has been updated there might be other effects before I can test.

Are there any other gryo type devices that also have temperature that I can use as template? No really sure how to return the temperature measure.

wilberforce:
@andycarle. 
https://github.com/Moddable-OpenSource/moddable/tree/public/examples/drivers/m5stickc-pedometer

Nice work! Love the example.

I'll play with the sensitivity value as I need to shake it get a step, otherwise looks good.

andycarle:
@wilberforce Excellent — I'm glad you the pedometer!

Can you verify that the m5stickc-imu example also works as expected on your device? I made quite a few changes to the driver to get it working on the device that I have, but I want to make sure they are compatible with the version that you have also.

wilberforce:
@andycarle
Your tweaks work great on my device. The smaller ball and action is much better.

Any thoughts on how the read of Temperature should be incorporated?

Should it be exposed in global scope?

andycarle:
@wilberforce 
Yes, I think that making a `temperature` global would be in the spirt of how we've been doing the M5 devices. I would probably just set it up the same as the accelerometer and gyro. Then applications can decide if they need the temperature sensor and how often it should poll and generate callbacks based on their application-specific needs.

andycarle:
@wilberforce I'm going to go ahead and close out this pull request since the relevant files have been merged in. Let's open up a clean issue or PR if you want to talk further on the temperature sensor.

PR #265, opened 2019-09-16
Adds AudioOut version of the piu/Sound example

zgramana:
Demonstrates some advanced capabilities of the `AudioOut` class,
as well as how to accomplish the same tasks as done with the
Sound class.

In addition, reproduces a device lockup issue when stopping an
audio playback stream by enqueuing `Flush`.


lprader:
Hi @zgramana—we'd be happy to accept this, but first we'll need you to send a PDF of a signed [CLA](https://github.com/Moddable-OpenSource/moddable/blob/78caaabe6ead2eac7ba82a20a8e755ba948f1e6d/licenses/Moddable%20CLA%20-%20individual.pdf) to [info@moddable.com](mailto:info@moddable.com). Thanks!

PR #283, opened 2019-11-04
Add mcconfig -n flag that skips flashing an attached device at the end of the build

jernst:
* This is hacky. (Best I could think of given generated Makefiles calling Makefiles.)
* Has not been tested on the PC or Mac (Linux only), however PC nmake changes have been made
* Works by overriding the flash command within the IDF Makefiles with a little python script.
* The python script records the parameters to the flash command that would have been used (sequence of address-binfile pairs). However, that file is currently below .../tmp/... in the IDF build hierarchy and has not been copied over into the Moddable output directory.

### DIFF
--- a/tools/mcconfig.js
+++ b/tools/mcconfig.js
@@ -1131,10 +1131,17 @@ export default class extends Tool {
 		file.generate(this);
 
 		if (this.make) {
-			if (this.windows)
-				this.then("nmake", "/nologo", "/f", path);
-			else
-				this.then("make", "-f", path);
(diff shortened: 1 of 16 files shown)


phoddie:
Great idea. We implemented this for all three build platforms using the `-t` option with options to clean, build only, deploy only, and connect to the deployed binary with xsbug. Details are in this [blog post](https://blog.moddable.com/blog/moddable-sdk-improvements-for-esp32-projects/)  under the heading "Execution of Individual Build Steps."

PR #306, opened 2019-12-28
linux CLI platform

dckc:
see also https://github.com/Moddable-OpenSource/moddable/pull/305


dckc:
oops; wrong repo; I meant agoric-labs/moddable

PR #312, opened 2020-01-11
Fixed bug in PKCS1_5 using undefined BigInt function

diogoviannaaraujo:
PKCS1_5 was using a undefined BigInt function called `fromByteArray` and it was crashing on use.
It has been switched to use `fromArrayBuffer`


diogoviannaaraujo:
Keeping a fork only for this small error,

I guess the function name was changed in the past but this one was missed.

diogoviannaaraujo:
This was fixed on a commit some days ago, closing.

PR #320, opened 2020-01-26
Add charlieplexing example

bterlson:
Happy to make any changes as I'm fairly new at this. Let me know!

### DIFF
--- a/contributed/charlieplexer/main.js
+++ b/contributed/charlieplexer/main.js
@@ -1,0 +1,31 @@
+class Charlieplexer {
+  constructor(pins) {
+    this.initializePins(pins);
+  }
+
+  initializePins(pins) {
+    this.pins = pins.map(p => new Digital(p, Digital.Input));
+  }
+
+  scan() {
+    for (let i = 0; i < this.pins.length - 1; i++) {
+      for (let j = i + 1; j < this.pins.length; j++) {
+        const p1 = this.pins[i];
+        const p2 = this.pins[j];
+        p1.mode(Digital.Output);
+        p2.mode(Digital.Output);
+
+        p1.write(1);
+        p2.write(0);
+        Timer.delay(100);
+
+        p1.write(0);
+        p2.write(1);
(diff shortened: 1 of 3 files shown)


contributed/charlieplexer/main.js
```diff
@@ -0,0 +1,31 @@
...
+
+  initializePins(pins) {
+    this.pins = pins.map(p => new Digital(p, Digital.Input));
+  }
+
+  scan() {
+    for (let i = 0; i < this.pins.length - 1; i++) {
+      for (let j = i + 1; j < this.pins.length; j++) {
+        const p1 = this.pins[i];
+        const p2 = this.pins[j];
```
phoddie, line 14: FWIW - if you expect a large number of pins, we would use a local for the length. XS doesn't have the optimizations of V8 and friends, so minimizing property look-ups helps.

```
   const pins = this.pins;
    for (let i = 0, length = pins.length; i < length - 1; i++) {
        for (let j = i + 1; j < length; j++) {
        const p1 = pins[i];
        const p2 = pins[j];
```

Here it the number is generally small, so no problem.

phoddie (approved):

contributed/charlieplexer/main.js
bterlson: @phoddie this is a wonderful point - I'm so used to v8-alike's optimizations it affects the way I code. Is there a document that goes over advice like this? I imagine there's a number of anti-patterns once you know which optimizations are present or not in xs.

phoddie:
Merged.

contributed/charlieplexer/main.js
phoddie: No document at the moment. It would make a good blog post. I've added it to the list on the wall!

A good rule is to think of XS more like a non-optimizing compiler: the code you write is close to what will execute. There are some optimizations, of course, but surely no runtime execution monitoring to optimize the JIT generated code. ;)

PR #323, opened 2020-01-28
export mxPromiseStatus() as fxPromiseIsPending(), fxPromiseIsRejected()

dckc:
This allows a platform to run an event loop until a promise returned from `main()` is resolved and to set the process exit code to non-zero if the promise was rejected.

Perhaps the 2 new functions should be combined?


phoddie:
I believe this PR was to support the original work on what has become the xs snapshot mechanism. As it isn't used there, I'm going to close this out. If I've misunderstand, please re-open with an explanation. Thanks.

dckc:
This was orthogonal to snapshots.

But closing it is reasonable; I think I have seen existing APIs that make this unnecessary.

PR #324, opened 2020-01-28
http async replies, upgrade hook

### LINKED ISSUE #308: how to upgrade an http server to websockets?
The websocket server seems to always create its own listener socket. I'm trying to support an application (Agoric ag-solo, [web.js][]) that initially uses its socket for ordinary GET requests for index.html and such before [UPGRADEing to websockets][2]. The popular [ws][] package supports a `noServer` mode where an http handler can pass its socket when an UPGRDADE request comes in.

Any suggestions on how to do this?

My guess is I should make a revised websockets.js module.

[web.js]: https://github.com/Agoric/agoric-sdk/blob/master/packages/cosmic-swingset/lib/ag-solo/web.js
[2]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#Upgrading_to_a_WebSocket_connection
[ws]: https://www.npmjs.com/package/ws

### LINKED ISSUE #313: websocket send() can't write messages > kTxBufferSize (1024 bytes)
websocket.js passes up to 64k of data to one call to socket.send(), but kTxBufferSize is only 1024 bytes, so I get:

```
modules/network/socket/lin/modSocket.c (761) # Exception: (host): can't write all data!
```

The docs say to use `socket.send()` with no args to get the amount of data that can be sent.

I guess I can bump up `kTxBufferSize` but websocket.js should break the data into pieces, no?

dckc:
fixes #308

doesn't fix https://github.com/Moddable-OpenSource/moddable/issues/313 but works around it to some extent

### DIFF
--- a/modules/network/http/http.js
+++ b/modules/network/http/http.js
@@ -596,53 +596,54 @@ function server(message, value, etc) {
 			let first;
 
 			if (7 === this.state) {
-				let response = this.callback(Server.prepareResponse);		// prepare response
+				let responseP = Promise.resolve(this.callback(Server.prepareResponse));		// prepare response
+				responseP.then(response => {
+					let parts = [];
+					let status = (!response || (undefined === response.status)) ? 200 : response.status;
+					let message = (!response || (undefined === response.reason)) ? reason(status) : response.reason.toString();
+					parts.push("HTTP/1.1 ", status.toString(), " ", message, "\r\n",
+										 "connection: ", "close\r\n");
 
-				let parts = [];
-				let status = (!response || (undefined === response.status)) ? 200 : response.status;
-				let message = (!response || (undefined === response.reason)) ? reason(status) : response.reason.toString();
-				parts.push("HTTP/1.1 ", status.toString(), " ", message, "\r\n",
-							"connection: ", "close\r\n");
+					if (response) {
+						let byteLength;
 
-				if (response) {
-					let byteLength;
-
-					for (let i = 0, headers = response.headers; headers && (i < headers.length); i += 2) {
-						parts.push(headers[i], ": ", headers[i + 1].toString(), "\r\n");
-						if ("content-length" == headers[i].toLowerCase())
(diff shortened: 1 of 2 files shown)


phoddie (commented):
I'm unsure about the goal of this change. It looks like it intentionally stalls-out the HTTP request after all the request headers have been received. How is that useful?

phoddie (commented):
By deign, the HTTP implementation does not use promises. We'll need to find another solution to the need this is addressing.

dckc:
> I'm unsure about the goal of this change. It looks like it intentionally stalls-out the HTTP request after all the request headers have been received. How is that useful?

As noted in #308, to upgrade an http server to websockets.

> By deign, the HTTP implementation does not use promises.

I think I was also emulating the node http API, where replies can be computed asynchronously.

Can you elaborate? What design constraint conflicts with the use of promises?

I can see a number of outcomes here:
   - you decide that what I've requested here is OK (seems unlikely)
   - Agoric maintains its own http module that supports async and websockets upgrade
   - Agoric drops some of its requirements
     - In recent developments, we've reduced the scope of what we're using XS for; we don't use http nor sockets at all. But we might expand the scope again.

phoddie:
I saw the note in #308 that the change is to support websocket upgrades. There is more going on which is not visible from the change. How does the proposed change support the websocket case you outline?

Regarding HTTP, the design constraint is specifically to not use Promises to eliminate unneeded overhead. Consequently, the PR cannot be accepted.

The potential solutions you enumerate above are: accept changes inconsistent with the implementation, fork, or change your goals. That overlooks my preferred solution - establish the requirements clearly, then design and implement a solution. Happy to work with you on that.

dckc:
Oops; of course; I left the best outcome out. rephrase:

It's been a while since I knew exactly why I (thought I) needed this. I hope I can explain it better in due course, but other work seems to have higher priority these days; I can double-check.

PR #326, opened 2020-01-29
Fix MQTT docs

bartmichu:
Well, the MQTT documentation is non-existent :) but I can fix the links. Ideally someone would create some examples.

### DIFF
--- a/documentation/network/network.md
+++ b/documentation/network/network.md
@@ -1342,6 +1342,6 @@ ping.close();
 <a id="mqtt"></a>
 ## class MQTT
 
-- **Source code:** [ping](../../modules/network/mqtt)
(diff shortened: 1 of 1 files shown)


andycarle:
@bartmichu Thank you for the fixes -- I'll get those merged in and you should see them in our next open source push. We'll also take a look at moving an MQTT example or two into the open source tree.

andycarle:
We've fleshed out the MQTT documentation a bit [here](https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/network/network.md#mqtt) and added [two new MQTT examples](https://github.com/Moddable-OpenSource/moddable/tree/public/examples/network/mqtt). Cheers!

PR #362, opened 2020-05-18
Fix broken XS Platform link

whymarrh:
This PR fixes the link to the _XS Platform_ document, which was malformed.

Note the extra `%` before this change.


phoddie:
Thank you! We'll merge that.

PR #394, opened 2020-07-27
Add M5Stick-C 80x160 screen

wilberforce:
image

This add as 80x160 and 160x8-0 screen to the simulator.

Has been tested on the win target - lin as been updated - but not tested. Was unsure on how to update the mac/wasm targets


phoddie:
merged

PR #397, opened 2020-08-07
Add e-paper 400x300 black/white/red configuration

meganetaaan:
Edz1qKvU0AEfRry

This PR adds configuration of an  e-paper that has 400x300 resolution and black/white/red colors.

* I tried [Waveshare E-Ink raw display module](https://www.waveshare.com/product/displays/e-paper/epaper-2/4.2inch-e-paper-b.htm). Not Crystalfont's one. But it works without flaw.
* I used display resolution to decide which color to use with respect to current implementation. But I think I'd better have an independent property of color scheme rather than begin decided by resolution.
* The defined width/height is now used as an argument for the register that sets the display resolution.
* Removed the assumption that the width of the display is within 256 pixels.


phoddie:
The changes look reasonable. Nice surprise that the driver more-or-less works as-is with the WaveShare display.

I agree that it is a bizarre to base the color mode of the display one the dimensions. That's an unfortunate artifact of how the driver evolved. It should be possible to add an explicit mechanism to select the color mode. I'll give that a try.

phoddie:
In the next push the color mode of the display will be decoupled from the display dimensions as you suggested. It is a small change:

```
#ifdef MODDEF_DESTM32S_COLORS
	#define EPAPER MODDEF_DESTM32S_COLORS
#elif (122 == MODDEF_DESTM32S_WIDTH) && (250 == MODDEF_DESTM32S_HEIGHT)
```

Then a manifest can state the color mode, for example here for Moddable Three:

````
    "defines": {
        "destm32s": {
            "width": 122,
            "height": 250,
            "colors": "kePaperBlackWhite",
            ...
```

phoddie:
Merged!

PR #422, opened 2020-08-14
Add pins/* type definition

meganetaaan:
This pull request adds a new type definition file for the `pins` package.

* Ported from my [types-moddable](https://github.com/meganetaaan/types-moddable) repo, modifying a bit to match the way it was written in Moddable.
* I've omitted documentation this time, but I think it's good to add JSDoc-style comments to class definitions and methods so developers can view the descriptions through completions like VSCode. In this case, how to avoid double management with markdowns under the `documents` folder is under consideration.


phoddie:
This looks great. We'll merge the PR in our next update. 

Mentioning @bmeck here in case he has feedback.

Regarding documentation, I support making the documentation more accessible. I'm not personally a fan of documentation in source code files (even type declaration files) as makes reading the code more difficult and is more challenging to maintain. As the primarily maintainer of much of the Moddable SDK documentation I'd prefer to remain in pure markdown. But, maybe there is a solution that works well enough for everyone.

bmeck:
> Regarding documentation, I support making the documentation more accessible. I'm not personally a fan of documentation in source code files (even type declaration files) as makes reading the code more difficult and is more challenging to maintain. As the primarily maintainer of much of the Moddable SDK documentation I'd prefer to remain in pure markdown. But, maybe there is a solution that works well enough for everyone.

Perhaps it would be a possible compromise to just put [`@see`](https://jsdoc.app/tags-see.html) in the declarations files. I essentially see them as header files since you can see the API signatures and docs in IDEs from them. That way you could still have links in the IDE.

phoddie:
The use of `@see` seems like a good place to start.

Much of the documentation is in the same repository as the type declarations. The examples for `@see` and `{@link}` show full URLs:

```
/**
 * @see {@link http://github.com}
 */
```

Is it possible to use a relative URL here, or do we always need to link to the GitHub repository?

phoddie:
Merged!

Issue #339, opened 2020-03-16
Support M5StickC with MPU6886


takeru:
IMU sensor module in M5StickC is changed SH200Q to MPU6886.
I will create PR.

https://github.com/m5stack/M5StickC/pull/35

Tiryoh:
Hi @takeru,
Is there any update on this issue?

wilberforce:
I now have an m5 stick C with the new imu. I'm looking to add a driver for this.

Tiryoh:
@wilberforce Great! I have [working on it](https://github.com/Tiryoh/moddable/tree/feature/add-mpu6886), but still testing. I'd be happy if anyone like you who knows more about moddable than I do make a PR.

phoddie:
@takeru: I took a quick look at the code. It seems pretty reasonable. I am unfamiliar with the IMU, so my feedback is about structure, not the IMU details.

One small point, it looks like you copied some code that does this:

```
throw "some error happened";
```

That is allowed. However, it is unconventional to throw a string. Typically an instance of `Error` is thrown:

```
throw new Error("some error happened");
```

The `reboot` method might more consistently be named `initialize` or, perhaps as in another sensor code, `enable`. It does not appear to restart the IMU, but reinitialize it.

Finally, integration is important. Some builds of the m5 stick c use the `SH200Q` and others use `MPU6886`. Ideally the host should support both without requiring the developer to configure it manually. Here's the current code which works well for development of the `MPU6886` driver:

```
	// state.accelerometerGyro = new SH200Q;
	state.accelerometerGyro = new MPU6886;
```

Here's a simple solution to support both:

```
	try {
		state.accelerometerGyro = new SH200Q;
	catch {
		state.accelerometerGyro = new MPU6886;
	}
```

The downside to this is that developers with the MPU6886 will experience an exception -- and a distracting visit to xsbug -- each time they start-up because of the exception. You could use preferences to remember the IMU on the device:

```
const imu = Preference.get("config", "imu");
if (!imu) {
	try {
		state.accelerometerGyro = new  SH200Q;
		Preference.set("config", "imu", "SH200Q");
	catch {
		state.accelerometerGyro = new MPU6886;
		Preference.set("config", "imu", "MPU6886");
	}
}
else if ("SH200Q" === imu)
	state.accelerometerGyro = new SH200Q;
else
	state.accelerometerGyro = new MPU6886;
```

[...]

PR #427, opened 2020-08-23
M5 imu support for SH200Q and MPU6886

wilberforce:
m5-Stick C IMU integration - see #339

The base manifest now has `imu` section

```
"config": {
		"screen": "st7735",
		"touch": "",
		"imu" : "MPU6886"
	},
```

To use on older hardware change to `"imu" : "SH200Q". I went this way as if defaulting to MPU8266 with try/catch the it crashes with a panic halt on the old hardware.


phoddie:
> defaulting to MPU8266 with try/catch the it crashes with a panic halt on the old hardware

That's very unexpected. Would you share the details?

wilberforce:
> > defaulting to MPU8266 with try/catch the it crashes with a panic halt on the old hardware
> 
> That's very unexpected. Would you share the details?

I could not repeat  - new update silently detects and works without changes on both versions of the M5 Stick C

phoddie:
Merged!

PR #430, opened 2020-08-30
Add target M5 atom - variants: lite/matrix/echo

wilberforce:
This pull request adds support for the M5 atom range:

m5-atom-family

- Atom lite (Button and RGB Led)
- Atom Matrix (5x5 neopixels, IMU 6886)
- Atom Echo (NS41423 Speaker)

Demos:
- Atom lite: rotate RGB colours on led, on button press flash white
- Atom Matrix: change RGB colour depending on orientation
- Atom Echo: play sound effects on button press

The SPM1423 microphone does not have a driver yet, or the IR led


phoddie:
Thanks for putting these ports together. Apologies for the delay in getting to this.

Everything looks good, but I'd rather not take the Resource name change until it covers all the cases. Would you mind adjusting the examples to remove that dependency?

wilberforce:
@phoddie 

> Everything looks good, but I'd rather not take the Resource name change until it covers all the cases. Would you mind adjusting the examples to remove that dependency?

Sure - what are the other cases the zip variant?

I think there have been some changes around qio and dio and flash speed I might need to add to the manifest too. (And m5_Stick C)

Shall I move the effects from `/ examples/drivers/m5atom-echo/wav` to `/examples/assets/sounds/effects` ?

phoddie:
Thanks. The resource name implementation ignores the archive. I'm not quite sure how to fix that without introducing an ugly hack. It'll sort itself out.

> I think there have been some changes around qio and dio and flash speed I might need to add to the manifest too. (And m5_Stick C)

For ESP8266, you can see the flash parameters in the [manifest](https://github.com/Moddable-OpenSource/moddable/blob/3499033dfee86288ffeb0f51eca245a95ac24bec/build/devices/esp/targets/moddable_one/manifest.json#L6-L8). But, for ESP32, I think it has to go in SDKCONFIG. You can have a partial SDKCONFIG for the target that only includes the lines needed to set the flash parameters. The manifest documentation has two sections about that -- setting the path to the [SDKCONFIG fragment](https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/tools/manifest.md#esp32-specific-environment-variables) and [how the fragments are merged](https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/tools/manifest.md#how-sdkconfig-files-are-processed) to create a full SDK CONFIG.

> Shall I move the effects from / examples/drivers/m5atom-echo/wav to /examples/assets/sounds/effects ?

That's a great idea. But... I just noticed... unless those sound effects have a Creative Commons license, maybe we should use others.

wilberforce:
@phoddie 

Ready to commit!

phoddie:
Thanks for making those changes. We'll get this merged.

(One day Star Wars will be in the public domain... though it could be a few centuries if Disney has their way....)

phoddie:
Merged!

PR #436, opened 2020-09-03
make servo example work out of the box

abparker:
I'm using ubuntu 16.04, so the `/Users/...` path in the manifest didn't work for me. I also found that `Servo.h` is actually capitalized in the esp8266 library.


phoddie:
[Merged](https://github.com/Moddable-OpenSource/moddable/commit/5639abb24b6d725554969dc0be5822edb54a4a08). Thank you!

PR #455, opened 2020-09-18
Add M5Stack CORE2

meganetaaan:
output-palette

This PR adds support for M5Stack CORE2, a new development board with touchscreen by M5Stack.

https://m5stack.com/products/m5stack-core2-esp32-iot-development-kit

Currently available features are:

* LCD(ILI9341)
* Touch screen(FT6336U)
  * compatible with the current FT6206 driver. Except a difference that the touch sensor of CORE2 protrudes from the LCD.
* 9-axis IMU(MPU6886)
* Power management(Axp192)
* Speaker
* Bibration motor

WIP:

* Microphone
* 8MB PSRAM
  * Is there any way straightforward to enable it using manifest?

These commits are created by me and [@jpzukin](https://github.com/jpzukin). I'm asking to send CLA.


phoddie:
Thank you! We will review. We have received the CLA from @jpzukin.

build/devices/esp32/targets/m5stack_core2/setup-target.js
```diff
@@ -0,0 +1,161 @@
...
+	global.speaker = new AudioOut({streams: 4});
+	if (config.startupSound) {
+		speaker.callback = function() {this.stop()};
+		speaker.enqueue(0, AudioOut.Samples, new Resource(config.startupSound));
+		speaker.enqueue(0, AudioOut.Callback, 0);
+		speaker.start();
+	}
+
+	// bibration
+	global.bibration = {
```
phoddie: vibration?

modules/drivers/axp192/axp192.js
```diff
@@ -17,18 +17,133 @@
  *   along with the Moddable SDK Runtime.  If not, see <http://www.gnu.org/licenses/>.
  *
  */
-import I2C from 'pins/i2c';
+import SMBus from 'pins/smbus';
 
-export default class AXP192 extends I2C {
+export const kCHG_100mA = 0
+
+export default class AXP192 extends SMBus {
```
phoddie: AXP192 is a general purpose component. These changes are specific to one device. It would be better to keep AXP192 general purpose and put the device specific details in a subclass.

modules/drivers/axp192/axp192.js
```diff
@@ -17,18 +17,133 @@
...
 	constructor(it) {
+		if (it.address == null) {
+			it.address = 0x34
+		}
 		super(it);
-		this.initialize();
+		if (it.onInit != null) {
```
phoddie: Once there is a subclass, then `onInit` is unnecessary as the subclass constructor can do this

modules/drivers/axp192/axp192.js
```diff
@@ -17,18 +17,133 @@
...
+			it.onInit.apply(this)
+	}
+	}
+
+	//set led state(GPIO high active,set 1 to enable amplifier)
+	setSpeakerEnable(state) {
```
phoddie: I'm not sure, but maybe it would be better to make these setters -- it seems more natural:

```js
axp.speakerEnable = 1;
```
It looks like it may be useful to have getters for some as well (can't quite tell).

modules/drivers/ft6206/modFt6206.c
```diff
@@ -178,15 +184,15 @@ void xs_FT6202_read(xsMachine *the)
 
 		// reflect
 		if (MODDEF_FT6206_FLIPX)
-			x = 240 - x;
+		  x = MODDEF_FT6206_WIDTH - x;
```
phoddie: (lost a tab)

modules/drivers/ft6206/modFt6206.c
```diff
@@ -178,15 +184,15 @@ void xs_FT6202_read(xsMachine *the)
...
 
 		if (MODDEF_FT6206_FLIPY)
-			y = 320 - y;
+			y = MODDEF_FT6206_HEIGHT - y;
 
 		// scale
-		if (240 != MODDEF_FT6206_WIDTH)
+		if (MODDEF_FT6206_FITX && 240 != MODDEF_FT6206_WIDTH)
```
phoddie: The usual style is to have parenthesis (even where order of operations doesn't require it:

```js
if (MODDEF_FT6206_FITX && (240 != MODDEF_FT6206_WIDTH))
```

build/devices/esp32/targets/m5stack_core2/manifest.json
```diff
@@ -0,0 +1,95 @@
...
+	"modules": {
+		"setup/target": "./setup-target",
+		"pins/audioout": "$(MODULES)/pins/i2s/*",
+		"pins/smbus": "$(MODULES)/pins/smbus/smbus",
+		"*": [
+			"$(MODULES)/drivers/mpu6886/*",
+		]
+	},
+	"preload": [
+		"monitor",
```
phoddie: This seems out of place because this manifest doesn't include the digital monitor.

phoddie (commented):
Looks like a great start. Thank you.

I cannot run the code but I read through it and left some comments. I hope they are helpful. If something is unclear, please let me know.

jpzukin:
Thank you for reviewing! 
We will update it with reference to the review results.

build/devices/esp32/targets/m5stack_core2/manifest.json
meganetaaan: Removed monitor from preload.

modules/drivers/ft6206/modFt6206.c
meganetaaan: Fixed according to your suggestion.

modules/drivers/ft6206/modFt6206.c
meganetaaan: Oops! Fixed.

build/devices/esp32/targets/m5stack_core2/setup-target.js
meganetaaan: Excuse my bad English X( fixed.

modules/drivers/axp192/axp192.js
meganetaaan: I've separated device-specific parts from AXP192 driver, confirming both of CORE2 and M5StickC build works.

modules/drivers/axp192/axp192.js
meganetaaan: Removed onInit function. Instead a constructor of each subclass is responsible for initialization.

modules/drivers/axp192/axp192.js
meganetaaan: I modified it to use getter/setter. I also removed the device specific words (speaker, LCD) and modified the code to represent the function of the power management module AXP192 itself.
For DCDC, LDO and auxiliary GPIO of AXP192, I have bundled the common operations into a file local class (is this acceptable?).

meganetaaan:
Thank you! Almost fixed. I would like you to re-review Axp192 codes.

phoddie:
> 8MB PSRAM
   Is there any way straightforward to enable it using manifest?

I haven't used PSRAM myself, but my understanding is that it is enabled through settings in SDKCONFIG. The manifest can reference an SDKCONFIG fragment.  You can see an example of that in this manifest:

https://github.com/Moddable-OpenSource/moddable/blob/54d8e94710ed6f8c310f046c4fbc83c967ccd346/contributed/modClock/manifest.json#L2-L5

The mechanism is documented [here](https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/tools/manifest.md#esp32-specific-environment-variables).

> Microphone

FYI - I have pending changes to the audioIn modules that get it working for both M5Stack Fire and M5Atom Echo. That work is probably a good starting point for M5Stack CORE2.

modules/drivers/axp192/axp192.js
```diff
@@ -17,34 +17,141 @@
...
+    this.#parent.writeByte(this.#register, data);
 	}
 }
+
+export default class AXP192 extends SMBus {
+  constructor(it) {
+    if (it.address == null) {
+      it.address = 0x34;
+    }
+    super(it);
```
phoddie: It is preferred to not modify the input argument, as it may be frozen or the caller may not expect changes. Here's a concise way to do that.

```js
super({address: 0x34, ...it)}
```

modules/drivers/axp192/axp192.js
phoddie: I think it is better, thank you. I don't know the component well enough to offer further feedback. We can always refine the AXP192 module further down the road.

modules/drivers/axp192/axp192.js
meganetaaan: Pushed a fix according above.

meganetaaan:
Thank you for reviewing. I pushed all my local fix. I think it is ready to merge.
And I'm glad to know about PSRAM and AudioIn. I'll see them later!

phoddie:
Thanks for the changes. I think we are in a good place to merge this. I want to review once more first.

phoddie:
Yes, this looks ready. We should be able to push the merge shortly.

phoddie:
This has been merged. Thank you!

Please note that there have been a bunch of changes clean-up the M5* targets, correct inconsistencies, etc.

PR #490, opened 2020-10-29
get isIps flag before SPI initialization

meganetaaan:
I found that `screen.js` of current Moddable build does not work on my old M5Stack.

This PR fixes the timing of getting is-IPS flag according to [the original driver implementation of M5Stack](https://github.com/m5stack/M5Stack/blob/5b4a619cb4c0122d0c847714200c370c2e77b3a9/src/utility/In_eSPI.cpp#L367).

I don't understand the root cause but getting the is-IPS flag shall be before sending initialization commands. I confirmed this change works on both of my old and new M5Stack.

### DIFF
--- a/build/devices/esp32/targets/m5stack/screen.js
+++ b/build/devices/esp32/targets/m5stack/screen.js
@@ -3,9 +3,9 @@ import Digital from "pins/digital";
 
 export default class extends ILI9341 {
 	constructor(dictionary) {
+		let isIps = Digital.read(33); // TFT_RST
 		super(dictionary);
-
-		if (Digital.read(33))		// TFT_RST
+		if (isIps)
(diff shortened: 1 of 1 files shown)


phoddie:
The change makes sense. The display driver initializes the reset (RST) pin to reset the device when it is initialized when super() is called. On return that point, the reset pin is configured as an output. The read then reinitialize the pin, but... that's ugly, at best. So, agreed that it is better to read the pin as early as possible. That better matches what the M5Stack native implementation does. Thanks for tracking this down. We'll get it merged.

phoddie:
This has been [merged](https://github.com/Moddable-OpenSource/moddable/commit/eae3293e99897d6a30da2bef499483e801be5d5c).

PR #505, opened 2020-11-30
Add get batteryVoltage() and get batteryCurrent() for APX192

stc1988:
* `get batteryVoltage()`
    * The value is used for calculating remaining battery.
* `get batteryCurrent()`
    * Returns positive value when battery charging and negative value when battery powered.
    * The value is used for detect whether battery charging or battery powered.

I confirmed these functions with M5stick-c and M5stack Core2


phoddie:
Looks good, thank you. We'll get this merged  in the coming days.

PR #510, opened 2020-12-09
Getting Started minor fix for $MODDABLE to %MODDABLE% for Windows

cmidgley:
Trivial update to the Getting Started doc - it had $MODDABLE in the Update section for Windows rather than %MODDABLE%

### DIFF
--- a/documentation/Moddable SDK - Getting Started.md
+++ b/documentation/Moddable SDK - Getting Started.md
@@ -248,14 +248,14 @@ The Moddable SDK tools are frequently updated with improvements and added functi
 1. Update your local clone of the [Moddable repository](https://github.com/Moddable-OpenSource/moddable). For instance, using `Git Bash`:
 
 	```text
-	cd $MODDABLE
+	cd %MODDABLE%
 	git pull
 	```
 
 	> Note that if you have any local changes to Moddable repository files, you may need to stash your changes and then reapply them after pulling:
 
 	```text
-	cd $MODDABLE
+	cd %MODDABLE%
 	git stash push
 	git pull
 	git stash pop


andycarle:
Hi @cmidgley,

Thank you for the suggestion! I'm not sure this is exactly the right change to make, but I agree that this portion of the instructions may be too confusing.

That step in the Getting Started guide assumes that the user is working in [Git Bash](https://gitforwindows.org/) ("For instance, using `Git Bash`:"). The Git Bash tool, annoyingly, does not accept the Windows-conventional `%VAR%` notation. It instead uses the Linux-conventional `$VAR`. 

I think those instructions date to a time before it was common to install the Git Command Prompt integration on Windows, so we assumed everyone was using Git Bash. Perhaps the right fix at this point is to remove all mentions of Git Bash and the `$` syntax from the Windows instructions, assuming that people are using the Command Prompt integration instead.

I personally still use Git Bash, but I'm curious what the rest of the world is up to. @cmidgley, what tool do you personally use for managing git repos at the command line on Windows?

&nbsp;- Andy

cmidgley:
I did not catch that - my bad!  I generally just use the "x86 Native Tools Command Prompt for VS 2019", or standard Windows CMD shell with environment vars.    I rarely use Git Bash, as I do almost all work in either the basic Windows shell, or in VS Code (no shell).   My preferred mode is to integrate everything into VS Code so I don't even need to use the shell at all.  I'm hoping to work out how to do that in the coming weeks (including Intellisense).  I believe others have succeeded but I've not seen a write up yet.

This can be closed, as it is clearly correct and I just skipped over the detailed instructions!  Appreciate the feedback.

PR #511, opened 2020-12-10
Adjust makefile system to allow for build-only (-b) option

cmidgley:
Added ability to build system for compiling and linking without running the simulator or flashing the firmware, as well as not starting xsbug.  By default the behavior is the same as before; adding the "-b" option to mcconfig / mcrun will "build-only".

This is implemented in mcmanifest.js by creating the BUILD_ONLY variable for makefiles and setting it to 1 when -b is selected.

Updated all makefiles to support this feature, for both mcconfig and mcrun.

My environment is Windows 10, with targets of ESP32 and ESP8266.  I do not currently have access to other hosts (linux or osx) nor the other target microcontrollers, so my testing is somewhat limited.  It would be wonderful if somebody who has those alternatives could give this a shot and report back if it works / doesn't break anything!


wilberforce:
Current `mcconfig` and `mcrun`  have a `-t` target option, I think this is doing the same as this pull request?

https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/tools/tools.md

Build Targets mcconfig takes an optional -t target argument to specify a build target. The options for the target are:

```
clean: removes build outputs for the app
build: builds the app, but does not deploy it
deploy: deploys a previously built app without rebuilding
xsbug: connect the xsbug debugger to a previously-deployed app
all: performs all steps (This is the default value, when the -t flag is omitted)
```

cmidgley:
Thanks - I had a gitter discussion with Peter and he recommended I do a pull to add this feature, and I wasn't aware that it already existed!  I just tried the `-t build` and there are a few problems with it, such as it still launches the debugger when building for debug, and mcrun does not build-only (it still tries to flash it).  But sounds like this is the wrong approach, so let's close this pull.  Thanks.

Issue #502, opened 2020-11-24
serial2xsbug on Windows sends incorrect time zone to target microcontroller


wilberforce:
**Build environment:** Windows
**Target device:** ESP32

**Description**
a. On the windows target, the correct timezone for New Zealand is set (+13)
For the ESP32 it is setting incorrect to -12
So the time set is 23 hours behind

b. For the windows target Time.tz and Time.dst are `undefined`

c. SNTP also seems to be be affected - so looks like it's the time setting for the correct offset that is the issue

**Steps to Reproduce**

main.js:

```
import Time from "time";

let d = new Date();
trace(`Set time: ${d.getTime()} ${d.toString()} tz:${Time.timezone} dst:${Time.dst}\n`);
let seconds = d.getTime() / 1000;
let read = seconds;
let s = new Date(read * 1000);
trace(`Get time: ${read} ${s.toString()}\n`)
```

manifest.json:

```
{
	"include": [
		"$(MODDABLE)/examples/manifest_base.json",
		"$(MODDABLE)/examples/manifest_net.json"
	],
	"modules": {
		"*": [
			"./main"
		]
	}
}
```

**Actual outputs**

Windows:

```
Set time: 1606251468556 Wed Nov 25 2020 09:57:48 GMT+1300 tz:undefined dst:undefined
Get time: 1606251468.556 Wed Nov 25 2020 09:57:48 GMT+1300
```

ESP32:

```
Set time: 1606251628313 Tue Nov 24 2020 09:00:28 GMT-1200 tz:-46800 dst:3600
Get time: 1606251628.313 Tue Nov 24 2020 09:00:28 GMT-1200
```

Chrome:
```
new Date(1606251628313);
Wed Nov 25 2020 10:00:28 GMT+1300 (New Zealand Daylight Time)
```

What you expect to happen

Dst to be reflected on the ESP32 target. Day to be correct
Time.dst and Time.tz to be defined on windows (linux/mac) target

Related:

https://github.com/Moddable-OpenSource/moddable/issues/498

phoddie:
Sorry, I'm a little mixed-up. The report says the target is ESP32 but there's a lot of information about Windows.

The simulator on Windows does not implement the `Time` status methods -- they are all [stubs](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/base/time/win/modTime.c). That's consistent with your note that they return `undefined`.

I'm not sure if/how that is related to the timezone being off by an hour on ESP32.

[...]

PR #514, opened 2020-12-11
Correct timezone for windows: serial2xsbug

wilberforce:
https://github.com/Moddable-OpenSource/moddable/issues/502


phoddie:
merged

PR #525, opened 2020-12-28
Add HX711

meganetaaan:
Hello. This is a spin-out from a private team project I contribute to. I'm glad if you merge it.
This pull request adds a driver for [HX711 Load Cell Amplifier](https://www.sparkfun.com/products/13879), which is suitable for making smart scales.

* It has two wire interface, clock and data, but not an I2C. I implemented it in C since its communication process requires microsecond speed.
* I tested it with M5Stack/M5StickC, [Mini Weight Unit(HX711)](https://m5stack.com/products/weight-sensor-unit) and a [load cell straight bar](https://akizukidenshi.com/catalog/g/gP-12034/)

### DIFF
--- a/examples/drivers/hx711/main.js
+++ b/examples/drivers/hx711/main.js
@@ -1,3 +1,17 @@
+/*
+ * Copyright (c) 2020 Shinya Ishikawa
+ *
+ *   This file is part of the Moddable SDK.
+ *
+ *   This work is licensed under the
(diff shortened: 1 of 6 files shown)


phoddie:
Looks good. The C code is very straightforward. We'll get this merged. Thank you!

phoddie:
merged!

PR #529, opened 2020-12-29
Linux Preferences module

cmidgley:
Simple implementation of a Preferences module using JS.  Not very memory efficient as it loads the preferences as JSON into memory, and then converts it to an object.  Manifest was modified so that the JS version can be used (without Preload, as I need a writable class static member) vs. the .c version used by the other platforms.


phoddie:
merged!

PR #533, opened 2020-12-31
Update tools.md

johnwargo:
Noticed a typo in the file, so I fixed it.  I also cleaned up the layout of the file for consistency sake (I found a couple of extra line breaks). Replaced JS with JavaScript everywhere (because newbies are not going to know what an arbitrary abbreviation is - besides, why abbreviate JavaScript when you have the space to spell it out completely). Also Updated the command line options for `mcconfig` to include submodules.


phoddie:
Thank you for the PR. We ask contributors to complete our [Contributor License Agreement](https://github.com/Moddable-OpenSource/moddable/blob/public/licenses/Moddable%20CLA%20-%20individual.pdf) before we accept pull requests.

> Replaced JS with JavaScript everywhere (because newbies are not going to know what an arbitrary abbreviation is...

JS is hardly "an arbitrary abbreviation." And this documentation is absolutely not for "newbies". This documentation is for advanced developers looking to customize the build to achieve results beyond common uses described in the Getting Started documents.

johnwargo:
I disagree, the documentation is absolutely for newbies - its for anyone who wants to read it (both newbies and experienced developers).  Perhaps my statement about people not knowing what JS means was incorrect, but it's documentation, you have as much space as you need to use full words instead of abbreviations. Both JS and JavaScript were used on the document, I simply tried to make it consistent.

phoddie:
merged

PR #535, opened 2020-12-31
Update tools.md

johnwargo:
GitHub moved away from using indentation to indicate code blocks, instead using the triple-tic approach which gives writers the ability to specify the code type for the code block. When properly set, GitHub then renders proper syntax highlighting in the browser - delivering a better experience for the reader.

I noticed that the file used both the old (indented) and new (triple-tic) approach to setting code blocks, so I updated the file to use the more modern approach everywhere.


phoddie:
merged

PR #562, opened 2021-02-13
set/get brightness for M5stack core2

stc1988:
Setting DC3 voltage from 2500 to 3300 can be used for change the LCD brightness.

file_000


andycarle:
Thank you for the contribution! We've pulled your commit and it will appear in the open source tree with our next push.

I've also added a constructor for `Host.Backlight` to the M5Stack Core2 `setup-target.js` that uses this new functionality you've added. That allows the M5Stack Core2 to work with the `piu/backlight` example without any changes to the example.

andycarle:
This is now [live](https://github.com/Moddable-OpenSource/moddable/commit/5cd3ac7f7a338f544560dc844f16e3437c0476d9). Thank you again for the contribution!

PR #613, opened 2021-03-21
Update blink to be platform neutral-2 - support pico without explicit pins

wilberforce:
Sorry - tried to to do both in one patch but it didn't go well.

### DIFF
--- a/examples/pins/blink/main.js
+++ b/examples/pins/blink/main.js
@@ -14,21 +14,14 @@
 
 import Timer from "timer";
 import Digital from "pins/digital";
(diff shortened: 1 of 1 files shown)


phoddie:
This isn't quite right:

```js
	if ( config.led.pin2 ) Digital.write(config.led.pin2, count & 1);
```

When `pin2` is 0, it will skip `Digital.write`. Better to check against `undefined`:

```js
	if ( undefined !== config.led.pin2 ) Digital.write(config.led.pin2, count & 1);
```

PR #632, opened 2021-04-18
Windows instructions use unix commands and delimiters

johnwargo:
Windows installation instructions referred to macOS/Linux commands for changing to the user profile directory. Also, some windows commands used forward slashes instead of Windows' backslashes.

### DIFF
--- a/documentation/devices/esp32.md
+++ b/documentation/devices/esp32.md
@@ -339,21 +339,21 @@ The Moddable SDK build for ESP32 currently uses ESP-IDF v4.2 and the CMake optio
 4. If you did not clone the ESP-IDF during the ESP-IDF Tools installation, create an `esp32` directory in your home folder, either from File Explorer or a terminal. For instance, in Git Bash:
 
     ```text
-    cd ~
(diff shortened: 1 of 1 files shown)


andycarle:
Hi @johnwargo,

Those portions of the instructions are intended to be run in Git Bash. ("For instance, in Git Bash..."). Git Bash uses Linux-style commands.

That said, it may be time to cut out the Git Bash bits of the instructions. Back when these were written, using Git Command Prompt integration was somewhat uncommon. But I think it is now the default option when installing Git for Windows? If so, we could probably safely cut over to using all Command Prompt instructions.

johnwargo:
@andycarle wow, I never even noticed that 'For instance, in Git Bash:' - I never, ever use git bash on Windows, I would never even consider doing that for Windows software installation. I know you probably have people who do, but boy I expect that's an edge case. 

Yes, of course, I know that git bash uses Linux-style commands. I use the git install option that gives me access to those commands in the Windows command prompt (`ls -l` anyone?)

PR #659, opened 2021-06-18
fix: fx_Function_prototype_bound was missing from snapshot callbacks

dckc:
Should it go at the end for compatibility? I don't think that matters to us @Agoric yet.

cc @phoddie @warner @FUDCo @kriskowal


phoddie:
Nice find!

PR #663, opened 2021-07-10
Improve type definition

meganetaaan:
* Separated `piu/Timeline` type definition from `piu/MC`. Otherwise it throws module-not-found error like below.

```
../extern/avatar/src/emoticon.ts:2:22 - error TS2792: Cannot find module 'piu/Timeline'. Did you mean to set the 'moduleResolution' option to 'node', or to add aliases to the 'paths' option?

2 import Timeline from 'piu/Timeline'
                       ~~~~~~~~~~~~~~
```

* Added type definition for contributed `serial` module. I understand these API will be gradually moved to which uses TC53 specification. But it is necessary for [my ongoing project](https://github.com/meganetaaan/stack-chan) to build. Will be pleased if you consider it.

### DIFF
--- a/typings/piu/MC.d.ts
+++ b/typings/piu/MC.d.ts
@@ -513,37 +513,6 @@ declare module "piu/MC" {
     new(behaviorData: any, dictionary: ScrollerDictionary): Scroller;
   }
 
-  interface Timeline {
-    duration: number;
-    fraction: number;
-    time: number;
-    from(
-      target: object,
-      fromProperties: object,
-      duration: number,
-      easing?: string,
-      delay?: number
-    ): Timeline;
-    on(
-      target: object,
-      onProperties: object,
-      duration: number,
-      easing?: number,
-      delay?: number
-    ): Timeline;
-    seekTo(time: number): void;
-    to(
-      target: object,
-      fromProperties: object,
-      duration: number,
-      easing?: string,
-      delay?: number
-    ): Timeline;
-  }
-  interface TimelineConstructor {
-    new(): Timeline
-  }
-
   global {
     const Skin: SkinConstructor
     const Texture: TextureConstructor
@@ -561,7 +530,6 @@ declare module "piu/MC" {
     const Label: LabelConstructor
     const Transition: TransitionConstructor
     const Text: TextConstructor
-    const Timeline: TimelineConstructor
   }
  }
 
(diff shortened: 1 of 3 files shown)


meganetaaan:
Hello, could you take a look?

phoddie:
There's no problem to take the changes related to Timeline. Thank you for taking care of splitting that out.

I'm not all that excited about the serial typings as I don't think we should be encouraging new code to be written that uses an API that is due to be deprecated. In looking at your repository, it doesn't appear to contain code that uses the Serial module, but you build it in. Am I correct in understanding that you intend for it to be used by mods? If that's the case, why not just switch over the the TC53 APIs? They work well on ESP32 (and many other places!).

meganetaaan:
Thank you for reviewing!

>  it doesn't appear to contain code that uses the Serial module

Sorry for my confusion. It does not include Serial but dependent [Serial servo](https://github.com/meganetaaan/moddable-rs30x/blob/master/src/rs30x.ts) module does.

>why not just switch over the the TC53 APIs? They work well on ESP32 (and many other places!).

I didn't know that the TC53 API was so readily available. So I'm going to add the type and use it in the module :) . I'd like to keep this PR open for pushing updates.

phoddie:
Thanks for the reference to the Serial servo. I missed that in my previous review.

> I didn't know that the TC53 API was so readily available. 

Yes. We are using it in some of our client projects already. It is working out very nicely.

> So I'm going to add the type and use it in the module :) . 

Excellent!

> I'd like to keep this PR open for pushing updates.

Makes sense.

It should be possible to use TC53 Serial for the servo. The biggest difference is that TC53 Serial is always asynchronous. That won't impact your write calls, but it will change how you read. It looks like the servo only reads for two reasons:

1. To clear the echo of the written command. This should be easy. You can just have a count of bytes to ignore when they are received asynchronously.
2. To read the result of the status command. That will require some change to the public `readStatus` API to support asynchronous delivery of the result (via a callback or Promise). Alternatively you could poll for the status, so it is always available to `readStatus`. Depending on how the servo is used that polling may already be happening at a higher level. I don't know.

Seems like an interesting little project to convert that over. If I can help, please let me know.

stc1988:
> * Separated piu/Timeline type definition from piu/MC. Otherwise it throws module-not-found error like below.
> ```
> ../extern/avatar/src/emoticon.ts:2:22 - error TS2792: Cannot find module 'piu/Timeline'. Did you mean to set the 'moduleResolution' option to 'node', or to add aliases to the 'paths' option?
> 
> 2 import Timeline from 'piu/Timeline'
>                        ~~~~~~~~~~~~~~
> ```

I have same issue too. Could you merge piuTimeline definition?

phoddie:
Sure, the change to split out Timeline will be in the next Moddable SDK update.

phoddie:
This has been merged.

PR #676, opened 2021-07-29
Allow ap mode in sim, correct station event order

wilberforce:
- The events are delayed so they can been seen (~10x slower than esp32). 
- The password is ignored and a connection is simulated.
- If the password is the string `invalid!` the connection fails.


wilberforce:
@phoddie please review

phoddie (approved):
This is a reasonable. The `mode` could be used more fully, which would help to catch usage errors earlier. Once we've merged this, I'll take a pass through to apply that.

wilberforce:
Closed with https://github.com/Moddable-OpenSource/moddable/commit/f09b6444e1ab6ebb46eed5c67e9df781f29823e0

PR #682, opened 2021-08-11
removing slashes in RTOS path

danhellem:
the \ in the RTOS paths are not needed and break when doing cut and paste


phoddie:
@danhellem - thanks for the PR. It looks reasonable. I don't remember why those slashes are even there. We'll check here before merging.

phoddie:
We'll take the PR in our next update, which should be later this week.

PR #685, opened 2021-08-16
Add CRC type definitions

stc1988:


phoddie:
Thank you for the PR. It may not be quite correct. The type definition requires that the input buffer be a `Uint8Array`. The CRC module accepts any of the following:

- `ArrayBuffer`
- Any `TypedArray`,including `Uint8Array`
- `HostBuffer`

There's no easy way, at least at the moment, to express that. We should probably eventually address that more broadly, for example in xs.d.ts. But, that path isn't so clear. A practical start could be to enumerate the most common types `ArrayBuffer | HostBuffer | Uint8Array`.

stc1988:
Thank you for reviewing.
Type definition of input buffer is not correct as you commented.

I don't have any idea to define `TypedArray` properly at the moment.
So I fixed it to you suggested.

PR #700, opened 2021-09-11
Add module builder for base64

kriskowal:
This adds a builder function for the base64 module to add the functions to global scope, as opposed to the paved path for using it as a module. This aids https://github.com/agoric-labs/xsnap-pub/pull/4

My C is rusty with a small R, so please review skeptically. In particular, I cargoed the `mxNoFunctionLength` pragma from the UTF-8 codec and know not whether it is necessary or even helpful. I was also unable to cult the `mxmc` patterns I observed there without incurring the wrath of the compiler, so I stuck with the simple incantations patterned off precedent in `xsnap`.

### DIFF
--- a/modules/data/base64/modBase64.c
+++ b/modules/data/base64/modBase64.c
@@ -39,8 +39,8 @@
 	modeled on FskStrB64Decode and FskStrB64Encode from KPR.
 */
 
-#include "xsPlatform.h"
 #include "xs.h"
+#include "xsPlatform.h"
 
 void xs_base64_encode(xsMachine *the)
 {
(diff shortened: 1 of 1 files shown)


phoddie:
The changes are fine, thanks. The use of `xsmc` is optional but does generate smaller code which we care about on microcontrollers. I'll take care of that change. (To use the `xsmc` macros you have to include `"xsmc.h"` which wasn't previously used in the base64 implementation.)

The `mxNoFunctionLength` is used as a way to distinguish a Moddable SDK build from a command line tool. It is a little too hacky. I don't think it is necessary either, as the linker dead strips the builder if it is unreferenced.

When the `ArrayBuffer` length is 0, is the expected behavior of `encode` to return an empty string? If an empty string is passed to `decode`, then it should return a 0 length `ArrayBuffer`? I have that change working locally and will commit if that's the expected behavior.

kriskowal:
Yes, that’s the expectation.

phoddie:
Despite the branch conflict message above from GitHub, this PR has been [merged](https://github.com/Moddable-OpenSource/moddable/commit/8883f696f3a9631ef9c2c05b62b5da8695bae1a8). There are a few [changes](https://github.com/Moddable-OpenSource/moddable/commit/960c30d9e13027d50728cf65991bc996878db73a) beyond that:

- Supports zero length buffers/strings (requires the latest xs)
- `decode` accepts TypedArrays in addition to ArrayBuffers
- Fixed dangling pointer in `decode`
- Switched over the xsmc* APIs (long overdue change)
- Eliminated use of `mxNoFunctionLength` - not needed here

kriskowal:
Thank you. All of these improvements are wonderful.

PR #719, opened 2021-10-05
Heltec Wifi Kit 32 OLED screen support

tribalmatt:
This adds support for the Heltec Wifi Kit 32 with OLED screen.  https://heltec.org/project/wifi-kit-32/

This creates the target build directory and needed config files for the Heltec WiFi kit 32 and OLED screen.  

- This required an update to the 1306 driver to support proper resets in i2c path.  The resets should work for both SPI and i2c.  I don't have an SPI 1306 device to test with that requires a reset (RST) pin.  
- The provider.js file is my best attempt with what I understood.  Please review.

### DIFF
--- a/build/devices/esp32/targets/heltec_wifi_kit_32/host/provider.js
+++ b/build/devices/esp32/targets/heltec_wifi_kit_32/host/provider.js
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2021  Moddable Tech, Inc.
+ *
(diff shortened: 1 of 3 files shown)


tribalmatt:
In the manifest.json in the sample you assisted with you put the color format in the config section with "format": "Gray256".  In my example first try I added it to the driver defines "ssd1306": {.  I think both work.  Is there a reason or convention in the SDK to use one over the other.

tribalmatt:
This has been merged.

PR #745, opened 2021-12-10
Ignore MPU6886 init error of m5stack_core2

meganetaaan:
The MPU6886 IMU on the M5Stack CORE2 is mounted on the bottom board and is removed when stacking other expansion boards. If it is removed, the initialization will fail. This pull request deals with the problem by going through an exception during initialization (as in the M5Stack Fire).
4_6d91fc3c-69ef-4755-894f-dbbf0c7b19bd_1200x1200
https://shop.m5stack.com/products/m5stack-core2-esp32-iot-development-kit?variant=35960244109476


mkellner:
Committed db8f973 on Dec 10.

PR #754, opened 2021-12-21
Add PAJ7620U2 gesture sensor and example

stc1988:
This is  [PAJ7620U2](https://www.pixart.com/products-detail/37/PAJ7620U2)  gesture recognition sensor.

Support 9 gestures:Up, Down, Left, Right, Forward, Backward, Clockwise, Anti-Clockwise, Wave.

### DIFF
--- a/examples/drivers/sensors/paj7620u2/main.js
+++ b/examples/drivers/sensors/paj7620u2/main.js
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2021  Moddable Tech, Inc.
+ *
+ *   This file is part of the Moddable SDK.
+ *
+ *   This work is licensed under the
+ *       Creative Commons Attribution 4.0 International License.
+ *   To view a copy of this license, visit
+ *       <http://creativecommons.org/licenses/by/4.0>.
+ *   or send a letter to Creative Commons, PO Box 1866,
+ *   Mountain View, CA 94042, USA.
+ *
+ */
+
+import Gesture from "embedded:sensor/Gesture/PAJ7620U2";
+import Timer from "timer";
+
+const sensor = new Gesture({
+  sensor: {
+    ...device.I2C.default,
+    io: device.io.SMBus
+  }
+});
+
+Timer.repeat(function () {
+  const sample = sensor.sample();
+  if (sample != "None") {
+    trace(`Gesture:${sample}\n`);
+  }
+}, 100);
--- a/examples/drivers/sensors/paj7620u2/main.js
+++ b/examples/drivers/sensors/paj7620u2/main.js
@@ -24,7 +24,7 @@ const sensor = new Gesture({
 
 Timer.repeat(function () {
   const sample = sensor.sample();
-  if (sample != "None") {
+  if (sample) {
     trace(`Gesture:${sample}\n`);
   }
 }, 100);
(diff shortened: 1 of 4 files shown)


phoddie:
Wow! This PR is great. Very cool to see the use of Ecma-419 and the effort to follow the Sensor Class Pattern.

I have a few suggestions to better follow the intent of Ecma-419.

- The `sample` method should return `undefined` when there is no value available. This is consistent with `read` in the IO Class Pattern. 
- `invertYAxis` and `invertXAxis` are implemented as toggles, but the Ecma-419 approach would be to make them properties with values of the object passed to  `configure`.
- The `configure` method is typically where `enable`, `invertYAxis`, and `invertXAxis` are implemented.
- The sensor should be initialized to a known state by the constructor. This gives the caller a predicable state. For example
```js
this.configure({enable: true, invertXAxis: false, invertYAxis: false});
```

This is a rough sketch of `configure` (I'm not sure the logic for setting the flags is correct in the original, so I'm  not sure about it here  either):

```js
configure(options) {
	this.#selectBank(1);

	if ("enabled" in options) {
		this.#io.writeByte(REGISTERS.OPERATION_ENABLE, Number(options.enabled));
	}

	if ("invertXAxis" in options) {
		let data = this.#io.readByte(REGISTERS.LENS_ORIENTATION);
		data ^= 1 << (options.invertAxis ? 0 : 1);
		this.#io.writeByte(REGISTERS.LENS_ORIENTATION, data);
	}
	if ("invertYAxis" in options) {
		let data = this.#io.readByte(REGISTERS.LENS_ORIENTATION);
		data ^= 1 << (options.invertYAxis ? 1 : 0);
		this.#io.writeByte(REGISTERS.LENS_ORIENTATION, data);
	}

	this.#selectBank(0);
}
```

And a few minor details:

- `Object.freeze(PAJ7620U2.prototype);` is not necessary any longer. The XS linker does this automatically when building for the Moddable SDK.
- The `invertAxis` function appears to be  a  no-op (`this.#invertAxis = this.#invertAxis`)
- It is great to see the use of private methods and fields. The `setGestureMode` method seems to be private but is publicly accessible.

phoddie:
The Ecma-419 version of the [FT6202 touch driver](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/drivers/sensors/ft6206/ft6206.js) is a good reference for this PAJ7620U2 driver.

-  When there is no sample ready, it [returns `undefined`](https://github.com/Moddable-OpenSource/moddable/blob/2f93df297782cf0a31c1797b246becbf5e546f4f/modules/drivers/sensors/ft6206/ft6206.js#L113-L115).
- It supports X and Y axis inversion with a single `flip` property that can have values of `"none"`, `"h"`, `"v"`, and `"hv"`.
- It has a [thorough implementation](https://github.com/Moddable-OpenSource/moddable/blob/2f93df297782cf0a31c1797b246becbf5e546f4f/modules/drivers/sensors/ft6206/ft6206.js#L69-L109) of `configure`
- It [calls `configure`](https://github.com/Moddable-OpenSource/moddable/blob/2f93df297782cf0a31c1797b246becbf5e546f4f/modules/drivers/sensors/ft6206/ft6206.js#L57-L61) from the constructor to initialize itself to a known state.

There's also a helpful [software data sheet](https://ecmatc53.github.io/spec/web/Touch-FocalTech-FT6x06.html) available. We are trying to create these for each Ecma-419 supported component to give developers a consistent reference.

stc1988:
Thank you for reviewing. I learned much about Sensor Class Pattern.
I made some fix to follow Ecma-419.

phoddie:
@stc1988 – thank you for updating your code. It looks very good. We'll work on merging the PR.

>  I learned much about Sensor Class Pattern.

All of us are still learning. ;)

PR #824, opened 2022-01-29
calculate correct voltage value, fixes #823

### LINKED ISSUE #823: esp32/m5paper wrong battery voltage calculation
**Build environment:** macOS
**Target device:** esp32/m5paper

**Description**
Voltage read by the `device.peripheral.battery.Default` class are incorrect.  

**Steps to Reproduce**
1. Compile this sample code: 
```js
const battery = new globalThis.device.peripheral.battery.Default();
trace(`Voltage is: ${battery.read()}\n`);
```
2. See a value of ~`10,xxxx` being reported for a nearly fully charged battery. 

**Other information**
It looks like this is the problem: 
https://github.com/Moddable-OpenSource/moddable/blob/e55fbbf8c9a450acb052fc1630171f04a11c25c2/build/devices/esp32/targets/m5paper/host/provider.js#L205-L209

I guess this comes from the m5ink implementation but the factory implementations are different. See 
https://github.com/m5stack/M5EPD/blob/bf4bd28f690d30a01310c8f538761f3b070b77f3/src/M5EPD.cpp#L107-L118 and 
https://github.com/m5stack/M5-CoreInk/blob/47951a3a8787e0fbaa999e06ec2d0b55a6583af7/examples/FactoryTest/FactoryTest.ino#L52-L62
respectively. 

After updating the implementation to
```js
let value = this.#analog.read() * 3300;
return value / (1 << this.#analog.resolution) / 0.5;
```
the values look much more reasonable. On thing to not though is that the M5EPD driver is [initialising the ADC_UNIT_1 with a different reference voltage](https://github.com/m5stack/M5EPD/blob/bf4bd28f690d30a01310c8f538761f3b070b77f3/src/M5EPD.cpp#L93) (3600 instead of [V_REF 1100 in the esp32 io module](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/io/analog/esp32/_analog.c#L149)).

FWeinb:
See #823 for more information.

### DIFF
--- a/build/devices/esp32/targets/m5paper/host/provider.js
+++ b/build/devices/esp32/targets/m5paper/host/provider.js
@@ -204,8 +204,7 @@ const device = {
 				}
 				read() {
 					let value = this.#analog.read() * 3300;
-					value *= 25.1 / 5.1 / 1000;
-					return value / (1 << this.#analog.resolution);
+					return value / (1 << this.#analog.resolution) / 0.5;
 				}
 			}
 		}


phoddie:
This has been merged.

PR #827, opened 2022-01-30
Add support for m5paper in screen test and mcsim

FWeinb:
Added support for the m5paper to the screen test (windows/linux builds are untested) and the mcsim application. I also added basic button support to the flashcard example to show the button simulation: 

https://user-images.githubusercontent.com/1250430/151712144-cc9953a3-34f6-464d-b437-e3ffe7375096.mov

I have not looked into it but it would be great to be able to scale the simulated screen as with a resolution of `960x540` the preview might not fit without scaling onto the screen. That would involve changes to the mcsim application as far as I can tell. 

It is really nice to work with this codebase! Using the PIU framework for all these applications makes this stuff really intuitive to work with.


phoddie:
Very cool! Thank you. We'll get this merged.

Much appreciated that you took care of both `screen test` and `mcsim`. The integration of the buttons in `mcsim` is the kind of thing we hope to see much more of moving forward. 

> I have not looked into it but it would be great to be able to scale the simulated screen as with a resolution of 960x540 the preview might not fit without scaling onto the screen. That would involve changes to the mcsim application as far as I can tell.

The main challenge is that our graphics primitives (Poco) don't support scaling directly and so Piu mostly doesn't either.  It could be done: piuPC is not the identical to piuMC to allow for these kinds of differences.

> It is really nice to work with this codebase! Using the PIU framework for all these applications makes this stuff really intuitive to work with.

Glad you are having fun. Can't wait to see what you build. ;) Being able to implement the simulator with the same language and APIs as the device code definitely helps. And having a simulator makes iterating on UI and implementation ideas much faster.

mkellner:
I tried this patch on Linux and MacOS.
There is an exception on startup of `epaper-flashcards`.

image

FWeinb:
Are you sure you are starting it via `mcconfig -d -m -p sim/m5paper`? Otherwise the custom host config won't be loaded by the simulator. This `subplatform` concept gets a bit harry because the `manifest.json` must include a `mac`/`win`/`linux` declaration as the mcsim is still executing on the specific platform, just the `subplatform` is changed.

mkellner:
Ah, my mistake. It appears to work with `-p sim/m5paper`. Thanks.

phoddie:
This has been merged [here](https://github.com/Moddable-OpenSource/moddable/commit/00e66e2f4ad35f2d17ab15166e07971ef06e87af) and [here](https://github.com/Moddable-OpenSource/moddable/commit/fb7b2c28c2e5e05f1d1c16082d8826385246ca0c).

PR #834, opened 2022-02-07
Battery simulation for m5paper in mcsim

FWeinb:
Similar to #827 added support for the battery in mcsim. Demo:

https://user-images.githubusercontent.com/1250430/152782825-beb104ab-ef05-49dd-816c-7a556eb4b697.mov

### DIFF
--- a/build/simulators/m5paper/peripherals.js
+++ b/build/simulators/m5paper/peripherals.js
@@ -1,21 +1,21 @@
 import Control from 'control';
 
 export class Battery extends Control {
-  #voltage;
+  #battery;
 
   constructor(options) {
     super(options);
-    this.#voltage = 4300; // in mV
+    this.#battery = 1.0;
     this.postJSON({
-      battery: this.#voltage,
+      battery: this.#battery,
     });
   }
   read() {
-    return this.#voltage;
+    return this.#battery;
   }
   onJSON(json) {
     if ('battery' in json) {
-      this.#voltage = json.battery;
+      this.#battery = json.battery;
     }
   }
 }
--- a/build/simulators/m5paper/peripherals.js
+++ b/build/simulators/m5paper/peripherals.js
@@ -32,7 +32,7 @@ export class RTC extends Control {
   }
 
   enabled() {
-    return this.#config.enabeld;
+    return this.#config.enabled;
   }
 
   get time() {
--- a/build/simulators/m5paper/peripherals.js
+++ b/build/simulators/m5paper/peripherals.js
@@ -1,0 +1,56 @@
+import Control from 'control';
+
+export class Battery extends Control {
+  #voltage;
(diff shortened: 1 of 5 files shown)


phoddie:
Excellent. The battery is really valuable in the simulator, because testing it in real-time on the device takes **forever**. Thank you.

build/simulators/m5paper/peripherals.js
```diff
@@ -0,0 +1,56 @@
...
+  constructor(options) {
+    super(options);
+    this.#config = { time: undefined, enabled: true };
+    this.postJSON({
+      rtc: this.#config,
+    });
+  }
+
+  enabled() {
+    return this.#config.enabeld;
```
phoddie: typo – should be `enabled`

phoddie:
If I'm reading the code correctly, it looks like you also have enough of an RTC simulation in place that a project can get /set the RTC, though the RTC time doesn't progress beyond the time set. That is useful for developing set-up code.

phoddie:
Regarding `read` on the battery, the returned value should be from 0 to 1, so that it is independent of the particular battery. The implementation of `read` would be:

```js
  read() {
    return (this.#voltage - 3300) / 1000;
  }
```

If the voltage is useful, that be a separate call.

build/simulators/m5paper/peripherals.js
FWeinb: Thanks.

FWeinb:
> If I'm reading the code correctly, it looks like you also have enough of an RTC simulation in place that a project can get /set the RTC, though the RTC time doesn't progress beyond the time set. That is useful for developing set-up code.

I played with a RTC Simulation, forgot to remove this stuff in this PR. If you want I can add that to the UI. 

> Regarding read on the battery, the returned value should be from 0 to 1

I mirrored the implementation of the battery from here: 
https://github.com/Moddable-OpenSource/moddable/blob/4e39861fe5f00b5006393a0b42f84b64f17b5ea0/build/devices/esp32/targets/m5paper/host/provider.js#L197-L209

that one also give back the voltage.

phoddie:
So then, it was my mistake. Apologies for not noticing sooner. It is kind of great that implementing the simulator revealed a problem in the device code thought. ;) That being the case, we should update both to return a value from 0 to 1.

FWeinb:
I fixed the simulator and the m5paper implementation.

phoddie:
Much appreciated. We'll get that merged.

phoddie:
Merged. I also updated battery `read` on the device to read several samples and average, as the reading was quite noisy in my tests. I built a small app to see that. It is in [this gist](https://gist.github.com/phoddie/c7b8bf4917e163a7bc3ff3ab96ca949c), FWIW.

Issue #833, opened 2022-02-06
Enabling SPIRAM Support for m5paper results in no boot 


FWeinb:
**Build environment:** macOS
**Target device:** m5paper

**Description**
I tried enabling the SPIRAM on the m5paper with no success. As soon as `CONFIG_ESP32_SPIRAM_SUPPORT=y` is enabled in the sdkconfig the esp32 won't boot anymore. The m5paper is using a `ESP32-D0WDQ6-V3` which is configured to use the IO17 (CLK) and IO16 (CS). Looking at the [schematic this looks correct](https://m5stack.oss-cn-shenzhen.aliyuncs.com/resource/docs/schematic/Core/m5paper/M5_PAPER_SCH.pdf) (see ESPSRAM64H section). I don't really now how to debug this as I don't get any output on Serial anymore (even after increasing the log level to INFO for both the bootloader as the application). I tried disabling the [CONFIG_SPIRAM_BOOT_INIT](https://demo-dijiudu.readthedocs.io/en/latest/api-reference/kconfig.html#spiram-boot-init) and enabling [SPIRAM_ALLOW_STACK_EXTERNAL_MEMORY](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig.html#config-spiram-allow-stack-external-memory) but nothing worked. 

**Steps to Reproduce**
1. Change this to `y`
https://github.com/Moddable-OpenSource/moddable/blob/a4e0ec573b647af85ea2879e9dc7e22973a3634a/build/devices/esp32/targets/m5paper/sdkconfig/sdkconfig.defaults#L1
2. Build using `mconfig -d 
3. xsbug.app will never connect to the device

**Expected behavior**
It should be possible to use the available SPIRAM.

phoddie:
> It should be possible to use the available SPIRAM

Yes, it should. Clearly setting Espressif's basic configuration option is not enough.

Strictly speaking, I don't think this is a Moddable SDK bug. I would be very happy to learn how to solve this though.

FWeinb:
Had some time to investigate a little more. I did notice the default console was disabled via the `CONFIG_ESP_CONSOLE_UART_NONE` flag, thats why i wasn't getting any debug output. Enabling the default `CONFIG_ESP_CONSOLE_UART_DEFAULT=y` console I now see that the spi_flash is correctly initialised: 
```
I (22) psram: This chip is ESP32-D0WD
[...]

PR #842, opened 2022-02-12
Add support for psram to m5paper target

FWeinb:
As extensively discussed in #833 and with awesome support by @phoddie this PR adds support for access to the PSRAM for the m5paper target. 

I tried to add the minimal amount of custom SDK settings here. Using `CONFIG_ESP32_REV_MIN_3` ensures that various workarounds won't be enabled, that way it is possible that future SDK Revision can make appropriate changes. 

Setting `DEBUGGER_SPEED: 115200` is currently used as a workaround because driving the UART Bus faster causes the bus to be corrupted.


phoddie:
This looks great. We'll get it merged. Thank you. This may be a helpful example for others looking to support PSRAM on other development boards.

phoddie:
This has been merged.

PR #856, opened 2022-02-23
m5StickC mcsim support

wilberforce:
Not sure why, but I had to change this to 0000 for it to appear:

`sortingTitle: '0000',`

The buttons don't seem to trigger the events (same for M5 paper). I'm not sure how the led should be done - just mirrored existing.


wilberforce:
`mcconfig -d -m -p sim/m5StickC`

image

mkellner:
Hello. It doesn't look right on my machine.

image

The skin file looks like it is the right size, but it appears to be getting scaled in mcsim.

wilberforce:
@mkellner 
> Hello. It doesn't look right on my machine.

I'm building on windows and it's fine for me...

`C:\Users\rhys\Projects\moddable\examples\piu\balls>mcconfig -d -m -p sim/m5StickC`

image

In your image, 0.png the canvas is 160x80

on the right of the image, the canvas is: 160x80 it's offset 31x44 which is correct. The background image is 256x128
It should be 342x170.

I have no idea where the size 256x128 is coming from? Is something getting cached - I can't find anything else this size.

what happens if you change:
`sortingTitle: '0000',` to `sortingTitle: '0010',`

mkellner:
Changing sortingTitle makes no difference.

wilberforce:
@mkellner - what os are you building on?

>The skin file looks like it is the right size
The skin file is displayed 256x124 and should be 342x170 - i have no idea why this would occur ? @phoddie ?

What about other orientations?

phoddie:
I don't have a good explanation at the moment. I can try to look later if this isn't sorted out by then.

mkellner:
Interesting.

I see that the images that you've checked in are at 96 pixels-per-inch.

If I make the image at 72 pixels-per-inch it displays properly on Mac.

I first tried on Linux to see if I would encounter the same problem, but I see a different problem there. With the build line `mcconfig -d -m -p sim/m5stickC`, I get an error:

```
### Error: './m5stickc/manifest.json': manifest not found!
```

It appears that the subplatform is being lowercased by the build tools. This is not a problem on MacOS or Windows which typically use a case-insensitive filesystem, but it is a problem on Linux. So, I think we'll have to change the directory name to lower-case `m5stickc`.

wilberforce:
@mkellner 
I've just updated with lower case, and now the correct device gets selected too, and dpi to 72 below...

mkellner:
Thank you. Look for it in the next release!

PR #860, opened 2022-03-01
qrcode piu example

wilberforce:


phoddie:
Thank you, but already committed my example to the Moddable SDK repository earlier this afternoon. ;)

wilberforce:
oh poos.

phoddie:
Sorry about that. At least we both had the same great idea.

wilberforce:
> Sorry about that. At least we both had the same great idea.

I had done a git pull before my branch as well ;-)

PR #864, opened 2022-03-02
Wemos OLED Lolin32 - Port

costa-victor:
## **Environment**
**Build environment:** Linux Mint 20 x86_64
**Target device:** Wemos OLED Lolin32
**Moddable version**: public - commit [e8efd64](https://github.com/costa-victor/moddable/commit/e8efd64e40bd7d1d0edd743fb88243d4157a97d1)
**ESPIDF version**:  v4.4

## **Description**
This is previously discussed [here](https://github.com/Moddable-OpenSource/moddable/discussions/863). The porting was based on Heltec WiFi Kit 32 target and it's very similar. The main differences are:

1. There is no built-in LED's in this board
2. SDA/SCL pins used for OLED display are different.

## **More information**
After testing a plenty of examples on this board after the porting, it's important to point out that:

1. Tested the button mapping and it's working.
2. Tested LED's examples.
3. LED's based examples crashes the application, as expected, because there is no LED pin declared for this target.
4. Tested OLED examples and it's working, as showed below:
wemos-oled-port


phoddie:
Congratulations! Very cool to see this working, especially so quickly.

Your code uses GPIO 21 & 22 for I²C while the [image](https://i0.wp.com/randomnerdtutorials.com/wp-content/uploads/2019/07/lolin32-oled-pinout.jpg) you link to for the pinouts uses GPIO 4 & 5. It is clearly working, so maybe it is the wrong image?

> backlight: 18, displayDC: 2, displaySelect: 15 are placeholders

Since you are using I²C to communicate with the display, `displayDC` and `displaySelect` are unused. You can simply remove them. If this board doesn't have a pin to control the backlight, then you can also remove `backlight`.

build/devices/esp32/targets/wemos_oled_lolin32/host/provider.js
```diff
@@ -0,0 +1,77 @@
...
+ *
+ *   The Moddable SDK Runtime is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU Lesser General Public License for more details.
+ *
+ *   You should have received a copy of the GNU Lesser General Public License
+ *   along with the Moddable SDK Runtime.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * FIXME: Analog port needs updating.
```
phoddie, line 19: The copyright block at the top should not be modified. Please move these notes to a separate comment below the copyright block. Thank you.

costa-victor:
Thanks Mr. @phoddie , i'm glad to hear that.

Regarding this point:
> Your code uses GPIO 21 & 22 for I²C while the [image](https://i0.wp.com/randomnerdtutorials.com/wp-content/uploads/2019/07/lolin32-oled-pinout.jpg) you link to for the pinouts uses GPIO 4 & 5. It is clearly working, so maybe it is the wrong image?

I followed the Heltec target example and it does the same, because the GPIO 21 & 22 are the default pins for I2C. I was in doubt about it, then I founded this at the [documentation](https://github.com/Moddable-OpenSource/moddable/blob/e9742b49d5132f2723d3a5fca45b65cf3818d01a/documentation/drivers/ssd1306/ssd1306.md#using-i2c).
> If not present, they may be provided by the constructor as the scl and sda properties of the dictionary. If they are not passed to the constructor, the default I2C interface is used.

So, I think it's correct and it works without changing the default I2C pins.

build/devices/esp32/targets/wemos_oled_lolin32/host/provider.js
costa-victor: Oh, got it. I will fix this. Thanks.

phoddie:
Regarding I²C, I understand now: you are leaving the default  I²C bus on GPIO 21 & 22 and using GPIO 4 & 5 for the display's I²C bus. That's fine.

costa-victor:
Closing because this has already been authorized.

Issue #871, opened 2022-03-13
Setup modules not loaded when running in a simulator


FWeinb:
**Build environment:** macOS
**Target device:** Simulator/m5paper

**Description**
I want to add support for the [`global.power`](https://github.com/Moddable-OpenSource/moddable/blob/5df3b556aa18cdefc03fd076b735ac26b4178aeb/build/devices/esp32/targets/m5paper/setup-target.js#L6-L19) object to the simulator but as the [`Control`-class relies on `screen`](https://github.com/Moddable-OpenSource/moddable/blob/5df3b556aa18cdefc03fd076b735ac26b4178aeb/build/simulators/modules/control.js#L14) I need to execute this during setup. But when adding a `setup/target` module to the `preload`-object in the [manifest.json of the m5paper simulator](https://github.com/Moddable-OpenSource/moddable/blob/public/build/simulators/m5paper/manifest.json) like this:
```js
{
  "modules": {
    	// ...
		"setup/target": "./setup-target"
  },
  "preload": [
 	//  ...
    "setup/target"
  ],
  // ...
}
```
with `setup-target.js`
```js
export default function () {
  trace(`Setup-Target called\n`);
  done();
}
```
the default exported function is never called.  

**Steps to Reproduce**
1. Copy the content of [these files](https://gist.github.com/FWeinb/a83e2a30e26cea0c01ceb4902f3bb35a) to `$MODDABLE/build/simulators/m5paper`
2. Run any of the `$MODDABLE/examples/piu/epaper-*` examples using `mcconfig -d -m -p sim/m5paper`  
3. See that `Setup-Target called` is never printed in the xsbug output. 

**Other information** [Edit]
I had a little more time to investigate this. It looks like the behaviour described in the [setup documentation](https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/base/setup.md) is only implemented for the embedded platforms in `$MODDABLE/xs/platforms/*`. But as mods are even working in the simulator I am not quite sure if I am on the right track here. 

Okay, it looks like you are building a shared library that exports the `fxScreenLaunch` function for the use in the simulator with the entry point [here in screen.c](https://github.com/Moddable-OpenSource/moddable/blob/public/build/simulators/modules/screen.c#L310) i think that would be the place to add the "setup/" init like done for the xsHost on an mc. 

Got it working. Here is an initial patch for the Simulator-`screen.c`:
```patch
[...]

PR #873, opened 2022-03-16
add support for "setup/*" modules in simulators

FWeinb:
As described in #871 this PR adds support for loading "setup/*" modules [as described here](https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/base/setup.md).


phoddie:
This has been merged.

PR #874, opened 2022-03-16
fix label for buttons in ButtonsRow for mcsim

FWeinb:
Currently the buttons in the ButtonsRow are using the parents label caused by a typo `$` instead of `$$`
Current behaviour:
three buttons with the label of the parent

With this PR applied:
three buttons with the correct label applied


phoddie:
Good find. Thank you. Embarrassingly, I noticed that recently, but didn't think to look into it....

phoddie:
This has been merged.

PR #893, opened 2022-04-01
Update pico.md to set PICO_GCC_ROOT on macOS

HipsterBrown:
More details about discovering this fix here: https://github.com/HipsterBrown/xs-dev/pull/42


mkellner:
This appears to have been fixed some time ago.

PR #901, opened 2022-04-16
typings for BLE modules

chigix:


phoddie:
Thank you for the PR! It will be great to have TypeScript support or BLE. We'll review this in the coming days.

phoddie (commented):
What is the reason for removing onDisconnected here?

phoddie:
Overall this looks good. Before we can merge the PR, you need to provide an executed [contributor agreement](https://github.com/Moddable-OpenSource/moddable/tree/public/licenses#contributor-license-agreement). You email a signed copy to [info@moddable.com](mailto:info@moddable.com).

chigix:
@phoddie 
Exactly, it seems my miss that the `onDisconnected` method is still existing in `BLEClient`.
However according to the method definition in `BLEClient` class:

https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/ble/bleclient.js#L47

It seems that there is no argument received when `onDisconnected` invoked.
Can we that method to the declaration of `onDisconnect()`?

chigix:
> Overall this looks good. Before we can merge the PR, you need to provide an executed [contributor agreement](https://github.com/Moddable-OpenSource/moddable/tree/public/licenses#contributor-license-agreement). You email a signed copy to [info@moddable.com](mailto:info@moddable.com).

Replied just now.

documentation/network/ble/ble.md
```diff
@@ -15,7 +15,7 @@ This document describes the Moddable SDK Bluetooth Low Energy (BLE) modules. Bot
 * [Using BLE](#usingble)
 * [BLE Client](#bleclient)
 	* [Class BLEClient](#classbleclient)
-	* [Class Device](#classdevice)
+	* [Class Client](#classclient)
```
andycarle: Thank you for the contribution! It will be very helpful to everyone to have TypeScript support for BLE.

One question: what is the reasoning behind changing the name of the `Device` class to `Client`? The BLE devices described by this class can be either BLE Server or BLE Client devices. And in this context—devices being discovered by a BLE Client—they will almost always be BLE Server devices. (More broadly, the devices described by this class can be either BLE Peripherals or BLE Centrals and either GATT Servers or GATT Clients.)

I think I would prefer a different name here. If there's a conflict on `Device`, perhaps just `BLEDevice`?

chigix: Hi, exactly it is because I see the `Client` defined in GATT module:
https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/ble/gatt.js#L23
This client definition seems consistent with the object device intialization across both `BLEServer` and `BLEClient`:
https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/ble/bleserver.js#L141
https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/ble/bleclient.js#L110

I believe the `Client` here would be more matching the reality where the client class is defined.

Also, I cannot find where is the class called `Device` exactly defined in this code base, I thought that the writing of Device seems a typo regarding this definition missing in code.

But if there is exactly the `Device` defined as an abstract class which I didn't recognize correctly, please let me know.

chigix: Also if you prefer `BLEDevice` than `Client`, I also hope this class naming could be initially employed in the real implementation codes, so that it can become a reference to support the type definition in TypeScript.

chigix:
> What is the reason for removing onDisconnected here?

@phoddie

After checked the Moddable JavaScript SDK source code again, the class `Client` exactly has no method defines `onDisconnected` event listener.

https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/ble/gatt.js#L23

Instead, I pushed a new commit that fixes the `onDisconnected(device)`:

https://github.com/Moddable-OpenSource/moddable/pull/901/files#diff-c6f2ca436e4bc6ba0016b22f5caecfd6f3b917fba6eedd73c0e52254c21bcea8R287-R294

This definition can be referenced from:
https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/ble/bleclient.js#L116

documentation/network/ble/ble.md
andycarle: Ah, I understand now; thank you for the clarification. 

I do think that the `Client` class that you reference in `gatt.js` is misnamed. It has not previously been an issue (and I had not noticed it) because that class is generally only instantiated by internal Moddable SDK code and then presented to scripts as an instance of the abstract `Device` class.

You can see that in action in examples that use the `onConnected` callback. For example, while `bleclient.js` references `Client`:

```js
const client = new Client({ address, addressType, connection:params.connection, ble });
this.onConnected(client);
```
the `onConnected` callbacks in script generally look like (this example from `uartclient.js`:
```js
onConnected(device) {
	device.discoverPrimaryService(SERVICE_UUID);
}
```

At any rate, I now agree that (if anything) this is a flaw in `bleclient.js` and `gatt.js`, not in your proposed Typing. So, I'm fine with taking this as-is, and then I'll consider whether there's something to be fixed on our side.

phoddie:
Closing as this was merged back in May. Thanks again!

PR #906, opened 2022-05-15
Add support for HostBuffer on String.fromArrayBuffer

cmidgley:
This pull extends the buffer sanity test to include `XS_HOST_KIND` (inside of `fx_String_fromArrayBuffer`) so that it can accept either an `ArrayBuffer` or a `HostBuffer`.    

Background: I use `HostBuffer` to map ESP32 fast/slow memory, and then use CDV to map an efficient data structure on top of it.  CDV uses `String.fromArrayBuffer` for strings, which caused an error when the code ran that the buffer was not an `ArrayBuffer`.  This change has been verified to resolve that specific problem when using `HostBuffer`.

### DIFF
--- a/xs/sources/xsString.c
+++ b/xs/sources/xsString.c
@@ -350,11 +350,11 @@ void fx_String_fromArrayBuffer(txMachine* the)
 	slot = mxArgv(0);
 	if (slot->kind == XS_REFERENCE_KIND) {
 		slot = slot->value.reference->next;
-		if (slot && (slot->kind == XS_ARRAY_BUFFER_KIND))
+		if (slot && (slot->kind == XS_ARRAY_BUFFER_KIND || slot->kind == XS_HOST_KIND))
 			arrayBuffer = slot;
 	}
 	if (!arrayBuffer)
-		mxTypeError("argument is no ArrayBuffer instance");
+		mxTypeError("argument is no ArrayBuffer/HostBuffer instance");
 	bufferInfo = arrayBuffer->next;
 	limit = bufferInfo->value.bufferInfo.length;
 	offset = fxArgToByteLength(the, 1, 0);


phoddie:
The change works for the case you describe, but doesn't appear to be safe for general use.

-  The change has the potential to allow reading host data which the host object does not intend to make directly accessible to scripts. Only `XS_HOST_KIND` slots that have a next slot of kind `XS_BUFFER_INFO_KIND` are intended to be accessible data.  That's a key difference between host data created with `xsSetHostData` and `xsSetHostBuffer`.
- The code also depends on `arrayBuffer->value.arrayBuffer.address` working for a host buffer. It happens to work because in the [`host` part of the union](https://github.com/Moddable-OpenSource/moddable/blob/58f48111f53cfc0521aafc10951c3ebeba218770/xs/sources/xsAll.h#L264) the `address` field is first as in the [`arrayBuffer` part](https://github.com/Moddable-OpenSource/moddable/blob/58f48111f53cfc0521aafc10951c3ebeba218770/xs/sources/xsAll.h#L258) of the union.

Implementing this properly is some more work. It probably isn't that much code. The [implementation of `_xsmcGetBuffer`](https://github.com/Moddable-OpenSource/moddable/blob/58f48111f53cfc0521aafc10951c3ebeba218770/xs/sources/xsmc.c#L309-L318) is a useful reference point for that. It would also allow `SharedArrayBuffer` as an input (as they are a variation of a host buffer). 

There are a couple of alternatives to consider. The `TextDecoder` class implementation accepts most any type of buffer, and should work for this case. Alternatively a temporary copy of the data could be made, though that could be more work to get compileDataView to play along.

cmidgley:
What is your preferred approach - modify CDV to use `TextDecoder` or have `String.fromArrayBuffer` work with host buffers?  For me, doing CDV will be much easier...  in which case, I'll just close this pull and update the one pending on CDV.

phoddie:
Good question. I think it is reasonable for `String.fromArrayBuffer` to also accept a `SharedArrayBuffer`, which would also allow a host buffer to work. I can look into that.

cmidgley:
Just hit another one in my attempts to switch from `ArrayBuffer` to `SharedArrayBuffer` ... `Logical.xor` also dislikes `SharedArrayBuffer` (likely other Logical operations, but that is the only one I'm using ... from my modified Web Socket client so all messages from the wire allocate into system memory).

phoddie:
Logical.xor is very easy to update to work with most buffers. It needs to cut over to the new `xsmcGetBufferReadable ` as described in this [blog post](https://blog.moddable.com/blog/buffers/). I've been working through the modules to get that in place.

phoddie:
(Diverging from the focus of this discussion, but [changes](https://github.com/Moddable-OpenSource/moddable/commit/f656662c1ab638fb1ebca0dce90ddc833278817c) to `Logic.xor` and friends are in place)

phoddie:
Today's push contains an [update](https://github.com/Moddable-OpenSource/moddable/commit/28e1bc72b4e20bbeb0f1ced36289336b140ae64b) for `String.fromArrayBuffer` so it works with `SharedArrayBuffer` and host buffers as well. It throws if passed a host object that isn't explicitly tagged as a buffer.

cmidgley:
That's great news - I'll give it a try in a bit, and will close this out assuming all works out.

cmidgley:
Sorry for the long delay ... but I'm finally back on latest code and can confirm this fix works.  Thanks!

PR #909, opened 2022-05-23
New console/log features for xsbug

cmidgley:
This pull adds two features to `xsbug`:

1) Adds a new Preferences options "Show Exceptions" to show/hide exceptions from the console/log.  By default it is on, so no behavior change to xsbug by default.  When disabled, new inbound exceptions will not be included in the log.

2) Implements support for colorizing trace output in the console/log.  You can now add a `<info>`, `<warn>` or `<error>` tag at the start of a trace line, and the line will take on the appropriate color based on the `xsbug` theme (dark/light).  Readme updated accordingly.

Background:

I run a automated unit and stress tests.  These tests  might induce hundreds or thousands of exceptions, and lots of informational messages intertwined with errors/warnings.  Adding the exception filter reduces the clutter of the log output when I'm doing negative testing, and colors help identify and separate individual tests and help isolate errors.


cmidgley:
Few adjustments/additions:
- Fixed mouse hover on colorized log lines so not not hand (stay as ibeam) and not try to link to log lines.
- Changed the code color assignments in theme for dark so that it's more consistent with lite, and results in keeping red as the 'error' color for both dark and lite.
- Added check for type script (.ts) to allow colorizing TypeScript source code
- Minor changes to docs with updated screen snap.

cmidgley:
Any thoughts on if this is something you might consider merging?  It's been working great for me with no change to the API surface or debugger protocol.

phoddie:
Apologies for the silence. No one here has had a chance to dig into this. I'll try.

phoddie:
This is working nicely for me.

I'm a little concerned about enabling syntax coloring for TypeScript. That was certainly not planned for in [the code](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/piu/PC/Code/piuCodeJS.c) that does the parse for syntax coloring isn't expecting that. Perhaps TypeScript syntax is close enough that it generally works well enough?

Should we also add ".mjs" as an extension in CodeView.js where you added the TypeScript check? I think that still comes up in the npm universe.

cmidgley:
I find the color works good enough w/TypeScript.  Perhaps not perfect, but much better than everything in black!  

I've never used .mjs or .cjs w/Moddable (nor some of the other ones out there like .es6 or .jsx/.tsx, though those are necessarily compatible with Moddable without a pre-processor).  If people are running into it, perhaps we can add some of them but I think the Moddable world is really focused on .js and .ts.

phoddie:
OK, let's try with the TypeScript mapping. I have a feeling it is a slippery slope though. We'll see.

".mjs" is just a native ECMAScript module, as far as I understand. Moddable SDK hosts, [even on microcontrollers](https://github.com/Moddable-OpenSource/moddable/blob/b620dca3612efe5ae46c1e5cf9ef5ac26b363be7/xs/platforms/mc/xsHosts.c#L229), recognize it. I think we should add that one. On the other hand, ".cjs" is definitely not supported.

cmidgley:
.mjs has been added.

phoddie:
Cool, thanks. We will get this merged in the next day or so.

phoddie:
Merged! Closing this out. Thank you for the changes. I can totally see how this will help developers working with large xsbug logs.

PR #924, opened 2022-07-14
fix: remove esp32/esp32_thing default display

Templarian:
As discussed below the `esp32/esp32_thing` probably should look like `esp32/esp32_thing_plus` and not default a display or include any additional modules.

https://github.com/Moddable-OpenSource/moddable/discussions/921

### DIFF
--- a/build/devices/esp32/targets/esp32_thing/manifest.json
+++ b/build/devices/esp32/targets/esp32_thing/manifest.json
@@ -1,9 +1,6 @@
 {
-	"include": [
-		"$(MODDABLE)/modules/drivers/ili9341/manifest.json"
-	],
 	"config": {
-		"screen": "ili9341",
+		"screen": "",
 		"touch": ""
 	},
 	"defines": {
@@ -15,12 +12,6 @@
 			"miso_pin": 12,
 			"mosi_pin": 13,
 			"sck_pin": 14
-		},
-		"ili9341": {
-			"hz": 40000000,
-			"cs_pin": 15,
-			"dc_pin": 2,
-			"spi_port": "HSPI_HOST"
 		}
 	}
 }


Templarian:
@phoddie Would it make sense to move the defines to the `modules\drivers\ili9341\manifest.json` for each platform? For consistency in docs and pins.

```json 
"ili9341": {
	"cs_pin": 15,
	"dc_pin": 2,
	"spi_port": "HSPI_HOST"
}
```

I ordered a feather ESP32 in addition to the `thing` / `thing_plus` and a `"ili9341"` compatible display so that I can verify more as I work on some doc PRs. This way I can test across a few different ESP32 dev boards. Still working on my main project so might be a second until I have those PRs ready.

phoddie:
> Would it make sense to move the defines to the modules\drivers\ili9341\manifest.json for each platform? For consistency in docs and pins.

I'm not sure. The ILI9341 driver manifest is entirely platform independent. If the ILI9341 is built into a device, the configuration for that device is in the platform manifest (here's the example from [Moddable Two](https://github.com/Moddable-OpenSource/moddable/blob/86c6dff2cad47b2b72ad52e86de9eb8acfbcb8c0/build/devices/esp32/targets/moddable_two/manifest.json#L23-L39)). For devices without a built-in ILI9341 with one connected, the defines generally go in the project manifest.

> I ordered a feather ESP32 in addition to the thing / thing_plus and a "ili9341" compatible display so that I can verify more as I work on some doc PRs. This way I can test across a few different ESP32 dev boards. 

That's really great. Thank you!

Templarian:
Closing this one as it was merged. Thanks @mkellner!

Still learning a lot. Want to make sure I have a good grasp before continuing doc work. Just wired up joycon joysticks which were fairly straightforward analog inputs.

PR #937, opened 2022-09-04
fix typescript (mainly commodetto)

Templarian:
Mainly fixing commodetto.

- Adds `bitmap: Bitmap` to the `parseBMF`.
- Move `import` inside of `declare` statements.
- Add empty `index.d.ts` for all folders so `typeRoots` works.
- Fixed the formatting on these files for future devs.

Question... should `xs.d.ts` be in the `typings` folder?

```json
{
    "strict": true,
    "noImplicitAny": true,
}
```

I'll update this if I run into any others. Turning on `noImplicitAny` found most of these.


phoddie:
Thanks for the PR!

I don't really understand why the empty `index.d.ts` files are necessary. I read the TypeScript docs but that was a little abstract. Would you mind explaining the motivation?

Templarian:
@phoddie It looks like VS gives an error `TS2688` if a `index.d.ts` file isn't found. Everything works good, but I just wanted no errors to show up in VS Code's "Problems" panel. I'm not 100% sure, but there might be another way to fix it.

image

Templarian:
Tried to keep the tsconfig file very vanilla. It could be something missing here...

```json
{
    "compilerOptions": {
        "baseUrl": ".",
        "module": "ESNext",
        "target": "ESNext",
        "strict": true,
        "noImplicitAny": true,
        "moduleResolution": "Node",
        "lib": ["ESNext"],
        "paths": {
            "ui:*": [
                "ui/*"
            ]
        },
        "typeRoots": [
            "./typings"
        ],
        "experimentalDecorators": true
    }
}
```

phoddie:
(FYI – I accidentally merged this PR yesterday. We backed that out so we can work though this discussion)

phoddie:
We are using TypeScript in different ways. I just let `mcconfig` run `tsc` automatically to translate TypeScript to JavaScript. Would you mind explaining how you are configuring Visual Studio Code generate the error you show above? I recall from past discussions with other regularly users of TypeScript that is more than one way to set that up. I'd like to duplicate your situation to either understand why adding the `index.d.ts` files is the best solution or to be able to explore alternatives. Thank you!

Templarian:
Yeah, I'm only using the `tsconfig.json` for VS Code and copy the moddable typings & xs.d.ts to a `typings` folder (and I'm mapping some paths). The first thing that came up for the error was below and seemed to make the error go away.

https://github.com/microsoft/TypeScript/issues/27956#issuecomment-430849185

phoddie:
Apologies for being slow to get to back to this PR. Following your notes above with a little patience, I was able to see the problems that motivated the addition of the empty `index.d.ts` files. It is a little bizarre, but seems OK (50% of all work involving TypeScript seems to involve placating TypeScript...).

The only bit I don't understand is the addition of `poco` to `global` in `system.d.ts`. It isn't (generally) at global and it isn't part of ECMA-419 (which is what `system.d.ts` supports). We can just remove that after the merge though, so no big deal. But, do you happen to remember what motivating adding it?

Templarian:
@phoddie You can remove that `poco` global. I think I was treating it as a global since it gets defined by moddable (?).

I'm on/off coding with Moddable, but mostly doing hardware stuff at the moment, so less code. I'll get back into this again.

phoddie:
Cool, will do. FWIW – the Moddable SDK proper doesn't define a `poco` global, since there are plenty of uses of the SDK that don't involve a display or graphics. Plenty of examples in the SDK, mostly Commodetto, do define a top-level module variable named `poco` which looks and somewhat behaves the same:

```
const poco = new Poco(screen, {pixels: screen.width * 4});
```

phoddie:
And merged.

PR #947, opened 2022-10-23
Supported M5StickC Plus device

kitazaki:
Added m5stick_cplus target files.
IMG_2340

M5StickC is EOL, but we can obtain M5StickC Plus.
https://docs.m5stack.com/en/core/m5stickc
https://docs.m5stack.com/en/core/m5stickc_plus

### DIFF
--- a/examples/network/ble/test-client/tests.js
+++ b/examples/network/ble/test-client/tests.js
@@ -248,6 +248,22 @@ class AdvertisementProperties extends Test {
     }
 }
 
+class Whitelist extends Test {
+    startTest(device) {
+        device.discoverPrimaryService(HEART_RATE_SERVICE_UUID);
+    }
+    onServices(services) {
+        if (services?.[0]?.uuid?.equals(HEART_RATE_SERVICE_UUID))
+            services[0].discoverCharacteristic(uuid`2A38`);
+    }
+    onCharacteristics(characteristics) {
+        if (characteristics?.[0]?.uuid?.equals(uuid`2A38`)) {
+            this.char = characteristics[0];
+            this.char.writeWithoutResponse(Math.floor(Math.random() * 128) + 1);
+        }
+    }
+}
+
 export default [
     ConnectDisconnectTest,      //0
     ReadGAP,                    //1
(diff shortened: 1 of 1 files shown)


kitazaki:
specifies the width 136 otherwise 135 because of the known issue
https://github.com/Moddable-OpenSource/moddable/discussions/708
 
other register parameters are based on 
https://github.com/m5stack/M5StickC-Plus/blob/master/src/utility/ST7789_Init.h
https://github.com/m5stack/M5StickC-Plus/blob/master/src/utility/ST7789_Defines.h

kitazaki:
reversed the specified width and height with the register parameter (0x36: Memory Access Data Control) to avoid the know issue.

phoddie:
Thanks for the PR! This looks good.

It looks like the setup-target.js file is **nearly** identical to the $MODDABLE/build/devices/esp32/targets/m5stick_c/setup-target.js. The only difference is checking for the earlier IMU on the M5Stick C. To simplify ongoing maintenance,  the m5stick_cplus manifest could reference the m5stick_c version of setup-target.js instead.

```
"setup/target": "../m5stick_c/setup-target"
```

What do you think?

kitazaki:
@phoddie 
I think that is a good idea to reduce the maintenance cost.
I've confirmed a normal operation with m5stick_c setup_target.js
M5StickC and C Plus are almost same except a LED and a buzzer.

| | M5StickC | M5StickC Plus
|--|--|--|
LCD Controler | ST7735S | ST7789V2
LCD Size | 0.96 inch | 1.14 inch
Screen size | 80×160 | 135×240
6-axis IMU | SH200Q or MPU6886 | MPU6886
Microphone | SPM1423 | SPM1423
Buzzer | n/a | GPIO2
PMIC | AXP192 | AXP192
Battery | 95mAh/3.7V | 120mAh/3.7V
HAT Pin | G0, G36, G26 | G0, G25/G36, G26
Grove Pin | G32, G33 | G32, G33

kitazaki:
updated manifest.json to reference "../m5stick_c/setup-target"

PR #951, opened 2022-10-30
Add typings for `structuredClone` and `deepEqual`

meganetaaan:
Trick or treat! Here is an tiny PR to add typings for `structuredClone` and `deepEqual`. Tested to compile and work on my project [1](https://github.com/meganetaaan/stack-chan/blob/feature/outline_avatar/firmware/stackchan/face/face-renderer.ts#L212), [2](https://github.com/meganetaaan/stack-chan/blob/feature/outline_avatar/firmware/stackchan/robot.ts#L110).


phoddie:
Happy Halloween!

The typings look fine and we'll get those integrated. Thank you.

FWIW – `structuredClone` has an obscure [optional second argument](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) which is also supported in the Moddable SDK. MDN documents the argument as "transferables" though appears to be more of a general options object that happens to only have a single property, `transfer`, defined.

meganetaaan:
Thank you for your information. I have added optional `transferable` argument into the definition.
I wrote `transfer` property should accept only objects and reject primitives that will not make sense in this context. Glad if you take a look.

phoddie:
Thank you for the update! It looks about right (though I am not so good at TypeScript...).

phoddie:
This has been merged.

PR #954, opened 2022-11-03
Misc support for M5StickC Plus

stc1988:
This PR contains below fixes.
* Add `M5StickC Plus` to device documentation for ESP32, `manifest_mod` and `mcbundle`.
* Screen rotation is compatible with `M5Stick C`.
  * Set `270` to default rotation.
  *  Fix some `examples.


examples/drivers/m5stickc-pedometer/main.js
```diff
@@ -18,15 +18,16 @@ import Resource from "Resource";
...
 const SHAKEVALUE = 3;
 
-let pixelsOut = new ILI9341({});
-const width = pixelsOut.width;
-const height = pixelsOut.height;
-const render = new Poco(pixelsOut, {rotation: 90});
+const rotation = config.rotation;
+const width = (rotation / 90) % 2 ? screen.height: screen.width;
+const height =(rotation / 90) % 2 ? screen.width: screen.height;
+let render = new Poco(screen, {width, height, rotation});
```
phoddie: Poco does not use the `width` and `height` from the options object. It [takes](https://github.com/Moddable-OpenSource/moddable/blob/968fe02408be0343a64c2b627266cb85af316267/modules/commodetto/commodettoPoco.js#L44-L45) `width` and `height` directly from the `screen` instance and then [applies](https://github.com/Moddable-OpenSource/moddable/blob/968fe02408be0343a64c2b627266cb85af316267/modules/commodetto/commodettoPoco.c#L77) the rotation. So, I think this can be simplified. Would you trying something like this?

```js
 let render = new Poco(screen, {rotation: config.rotation});
 const width = render.width, height = render.height;
```

stc1988:
Thank you for your review.
I pushed a commit based on the proposal.

phoddie:
Looks good. Thank you!

phoddie:
Merged!

PR #957, opened 2022-11-07
Add support for M5Atom U

stc1988:
This PR adds support for [M5Atom U](https://docs.m5stack.com/en/core/atom_u).
* Based on M5Atom Lite
* Built-in USB-A interface
* Digital microphone

I confirmed this works with below examples.
* helloworld
* drivers/m5atom-lite
* pins/audioin/wavserver


phoddie:
Looks great! Your are getting good at this. What device do you plan to support next? ;)

PR #959, opened 2022-11-08
Avoids mixing debug and oss-fuzz flags

raphdev:
This PR should only modify oss-fuzz builds (when `OSSFUZZ` is set) and fixes coverage builds. 

Debug build C options are mixed with oss-fuzz injected ones. This can cause issues particularly with coverage builds as flags can clash -- consequently coverage reports do not show correct figures. 

`OSSFUZZ` builds should have all the necessary options injected via `CFLAGS` and `CXXFLAGS`, so we only use those.


raphdev:
This was merged in 7a831ee514bdae43233e2a264c4e505425a29cd3. Closing

PR #966, opened 2022-11-23
Add TextDecoder and TextEncoder type definition

stc1988:
I made this PR based on [@types/web](https://www.npmjs.com/package/@types/web), but modified to match [test of Moddable SDK](https://github.com/Moddable-OpenSource/moddable/tree/public/tests/modules/data/text) .

### DIFF
--- a/typings/text/encoder.d.ts
+++ b/typings/text/encoder.d.ts
@@ -1,0 +1,38 @@
+/*
+* Copyright (c) 2022 Satoshi Tanaka
+*
+*   This file is part of the Moddable SDK Tools.
+*
+*   The Moddable SDK Tools is free software: you can redistribute it and/or modify
+*   it under the terms of the GNU General Public License as published by
+*   the Free Software Foundation, either version 3 of the License, or
+*   (at your option) any later version.
+*
+*   The Moddable SDK Tools is distributed in the hope that it will be useful,
+*   but WITHOUT ANY WARRANTY; without even the implied warranty of
+*   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*   GNU General Public License for more details.
+*
+*   You should have received a copy of the GNU General Public License
+*   along with the Moddable SDK Tools.  If not, see <http://www.gnu.org/licenses/>.
+*
+*/
+
+declare module "text/encoder" {
+
+    interface TextEncoderEncodeIntoResult {
+        read?: number;
+        written?: number;
+    }
+    interface TextEncoder {
+        encode(input: string | number): Uint8Array;
+        encodeInto(source: string, destination: Uint8Array): TextEncoderEncodeIntoResult;
+    }
+
+    var TextEncoder: {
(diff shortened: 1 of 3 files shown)


typings/text/encoder.d.ts
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+declare module "text/encoder" {
+
+    interface TextEncoderEncodeIntoResult {
+        read?: number;
+        written?: number;
+    }
+    interface TextEncoder {
+        encode(input: string | number): Uint8Array;
```
phoddie: Is "number" correct here? The implementation expects a string.

typings/text/encoder.d.ts
```diff
@@ -0,0 +1,38 @@
...
+*   You should have received a copy of the GNU General Public License
+*   along with the Moddable SDK Tools.  If not, see <http://www.gnu.org/licenses/>.
+*
+*/
+
+declare module "text/encoder" {
+
+    interface TextEncoderEncodeIntoResult {
+        read?: number;
+        written?: number;
```
phoddie: `read` and `write` are always set in the result. I see that the TypeScript DOM declarations make them optional, but I'm not sure that is correct.

phoddie:
Thank you for preparing this PR. It's helpful to have these typings available. I noted a couple of minor points above. If you would review those and, if appropriate, correct then, then we can merge this.

By the way, I'm curious about what you are using the `TextDecoder` and `TextEncoder` for.

typings/text/encoder.d.ts
stc1988: I see [specification](https://encoding.spec.whatwg.org/#interface-textencoder) and check behavior  again, they doesn't seem to be optional. I will fix it.

typings/text/encoder.d.ts
stc1988: At first, I defined it as `string `,  but   `number` is set to `input` in [these tests](https://github.com/Moddable-OpenSource/moddable/blob/public/tests/modules/data/text/encoder/encode.js#L22-L23) .

phoddie: Good idea to check the tests. Note that the tests also pass a `BigInt`, which is different from a `number`.

The `encode` function expects a string. It coerces its input to a string. More-or-less, if something is not a string, then the `encode` calls `toString` on it. This is the same behavior as browsers implement for `encode` and the Web type declarations do not list ` number` as a type.

In TypeScript declarations, my understanding is an argument that is expected to be a string, even if other values are accepted by calling `toString`, is still declared as a `string` rather than enumerating all possible types that can be coerced (because that list is basically all types).

stc1988:
Thank you for reviewing.
I pushed the commits that were pointed out.

phoddie:
Perfect, thank you. We will get these merged.

BTW, I appreciate your patience in working through these small details. There is a style to writing TypeScript declarations that isn't always obvious (at least not to me!). I think it is important that we make the effort to get it right, so the declarations match the expectations of developers using them.

PR #969, opened 2022-11-28
Improve typings for `AudioOut` and `Resource`

meganetaaan:
Here I updated typings for `AudioOut` and `Resource`.

- `callbacks` property of `AudioOut` for setting callback functions for each stream
- `length` member function of `AudioOut` that returns the number of queued items
- `copy` option of `Resource#slice` to get `HostBuffer` object

afaik latter two are undocumented. I found them in the [wavstreamer](https://github.com/phoddie/soundstream/blob/main/http-stream/wavstreamer.js) sample implementation.

### DIFF
--- a/typings/pins/audioout.d.ts
+++ b/typings/pins/audioout.d.ts
@@ -67,7 +67,7 @@ declare module 'pins/audioout' {
     public static readonly Silence: MixerSilence
   }
 
-  type AudioOutCallback = (value?: number) => void
+  type AudioOutCallback = (value: number) => void
   class AudioOut extends Mixer {
       start(): void
       stop(): void
--- a/typings/pins/audioout.d.ts
+++ b/typings/pins/audioout.d.ts
@@ -55,6 +55,8 @@ declare module 'pins/audioout' {
     public enqueue(stream: number, kind: MixerSilence, samples: number): void
     
     public mix(samplesNeeded: number): HostBuffer
+
+    public length(stream: number): number
   
     public static readonly Samples: MixerSamples
     public static readonly Flush: MixerFlush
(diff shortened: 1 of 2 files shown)


typings/pins/audioout.d.ts
```diff
@@ -65,10 +67,12 @@ declare module 'pins/audioout' {
     public static readonly Silence: MixerSilence
   }
 
+  type AudioOutCallback = (value?: number) => void
```
phoddie: This is a small point. I'm not sure why the `value` argument is declared as optional here. The AudioOut implementation always passes the number argument. I'm still not very familiar with TypeScript though. Perhaps it is declared this way so that the a callback that does not use the `value` argument can be passed of the form `callback()`?

meganetaaan: Thanks for pointing that out. It is my mistake. I had overlooked that the second argument number of AudioOut.enque(AudioOut.callback) is [required](https://github.com/Moddable-OpenSource/moddable/blame/public/typings/pins/audioout.d.ts#L50). I pushed a fix.

>Perhaps it is declared this way so that the a callback that does not use the value argument can be passed of the form callback()?

This is actually achieved even if `value` argment is not optional. The below is a valid code.

```typescript
type FunctionThatTakesOneNumber = (value: number) => void
const myFunction: FunctionThatTakesOneNumber = (/* it doesn't take any argument */) => {  }
```

So even if the type definition stipulates that a function must have arguments, it is up to the user to decide whether or not to actually use them.

phoddie: OK, I understand. Functions may ignore input arguments but, when they use an argument, they must respect its type. Thank you for helping me find my way through TypeScript!

By the way, in applying the update I noticed that the second form of `mix` (recently added) is missing, so I will add that too. It is useful for optimizing off-line mixing, like some of the audio tests do.

```ts
public mix(buffer: BufferLike): void
```

phoddie:
merged

PR #995, opened 2022-12-17
Add ResourceStreamer

meganetaaan:
I implemented `ResourceStreamer` class by adding `WavStreamer` compatible I/F to @phoddie 's [`resource-stream`  example](https://github.com/phoddie/soundstream/tree/main/resource-stream).

This is a local resource version of [`WavStreamer`](https://github.com/Moddable-OpenSource/moddable/blob/public/examples/pins/audioout/http-stream/wavstreamer.js).
Streaming is not necessary for local audio playback, but I think `ResourceStreamer` is convenient for interoperability with other streaming classes that have `onPlayed` callback and etc.

https://github.com/phoddie/soundstream/issues/1

### DIFF
--- a/examples/pins/audioout/resource-stream/main.js
+++ b/examples/pins/audioout/resource-stream/main.js
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2022  Moddable Tech, Inc.
+ *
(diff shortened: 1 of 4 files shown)


phoddie:
merged

PR #998, opened 2022-12-30
Sensor Overview

Sineos:
Since I did not find an overview of the supported sensors, I decided to create one.

Some things I noticed:
* https://github.com/Moddable-OpenSource/moddable/tree/public/modules/drivers and https://github.com/Moddable-OpenSource/moddable/tree/public/modules/drivers/sensors partially contain duplicate entries albeit with slight differences
* Same goes for https://github.com/Moddable-OpenSource/moddable/tree/public/modules/drivers/peripherals and https://github.com/Moddable-OpenSource/moddable/tree/public/modules/drivers/rtc

Hope this can be accepted without CLA since it is just leg work and no intellectual work.

Edit: The list as [spreadsheet](https://docs.google.com/spreadsheets/d/1YXuwi43OAS04rIZ8zkwXS5t_kHhf0b5I80Pif-WjB5I/edit?usp=sharing)


phoddie:
Thank you! This looks really valuable. I will need a little time to review. Thanks for your patience.

phoddie:
> Hope this can be accepted without CLA since it is just leg work and no intellectual work

We do prefer to have a CLA, but I agree that here it is not essential. I understand from your PR and comments that you are fine with us including this as part of the Moddable SDK. We'll get this merged.

FWIW – regarding your observation about duplication, that is a consequence of our move to using ECMA-419 standards wherever possible. The ECMA-419 sensor implementations are in `modules/drivers/sensors` and the ECMA-419 RTC implementations are in `modules/drivers/peripherals`. We clearly have some work to do to organize the repository to make this more clear.

Sineos:
> I understand from your PR and comments that you are fine with us including this as part of the Moddable SDK.

Absolutely. Alas I'm no real (not even an unreal one) developer, so I'm happy if I can give back at least a little to such great OSS projects.

> FWIW – regarding your observation about duplication, that is a consequence of our move to using ECMA-419 standards wherever possible. The ECMA-419 sensor implementations are in `modules/drivers/sensors` and the ECMA-419 RTC implementations are in `modules/drivers/peripherals`. We clearly have some work to do to organize the repository to make this more clear.

Thanks for the clarification. In a Node Red context, does this mean only the sensors in `modules/drivers/sensors` are actually usable? Would it make sense to add a small note to this new readme.md here about this?

phoddie:
> In a Node Red context, does this mean only the sensors in modules/drivers/sensors are actually usable?

Correct.

> Would it make sense to add a small note to this new readme.md here about this?

It is absolutely worth noting which sensor drivers conform to the ECMA-419 Sensor class.

Sineos:
Already went ahead. See https://github.com/Moddable-OpenSource/moddable/pull/1002

phoddie:
Cool, thanks. Clearly I'm still catching up!. I'll close this one out then.

PR #1004, opened 2023-01-13
Add metering in oss-fuzz harness

raphdev:
This PR modifies the OSS-fuzz harness so that the XS metering limits execution. By exiting gracefully (rather than being interrupted by a timeout) when reaching a computation limit we avoid the fuzzer interpreting long-running test cases (infinite or near-infinite loops) as a bug.

### DIFF
--- a/xs/tools/xst.c
+++ b/xs/tools/xst.c
@@ -1883,8 +1883,15 @@ int fuzz(int argc, char* argv[])
 #if OSSFUZZ
 static xsBooleanValue xsWithinComputeLimit(xsMachine* machine, xsUnsignedValue index)
 {
-	// some test262 fail at 2147483800
-	if (index > 1000000) {
+	#ifdef mxFuzzMeter
+		static const xsUnsignedValue mxFuzzMeterLimit = mxFuzzMeter;
+	#elif
+		// highest rate for test262 corpus was 2147483800
+		static const xsUnsignedValue mxFuzzMeterLimit = 214748380;
+	#endif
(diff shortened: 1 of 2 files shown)


xs/tools/xst.c
```diff
@@ -154,6 +154,36 @@ static int main262(int argc, char* argv[]);
...
+		(_THE)->scope = __HOST_JUMP__.scope, \
+		(_THE)->frame = __HOST_JUMP__.frame, \
+		(_THE)->code = __HOST_JUMP__.code, \
+		(_THE)->firstJump = __HOST_JUMP__.nextJump; \
+		break; \
+	} while(1)
+#else
+	#define xsBeginMetering(_THE, _CALLBACK, _STEP)
+	#define xsEndMetering(_THE)
+#endif
```
raphdev, line 186: I copied this from xsnap as current xs does not have these macros. Is there a better place for these to be defined?

phoddie: We'll take a look at whether it makes sense at this point to migrate the metering macros into xs.h. They were embedded in xsnap initially as it was the only place they were used.

raphdev:
As suggested, metering limit can be configured. I opted for a compile-time limit that can be set in oss-fuzz.

In oss-fuzz, https://github.com/google/oss-fuzz/pull/9458 sets the metering limit.

raphdev:
This was already merged in 5b8a037ac94f97e162f86548cc625f810527a6a6. Closing.

PR #1015, opened 2023-01-28
Add blank typings/text/index.d.ts, fixes #1014

### LINKED ISSUE #1014: typings/text is lacking index.d.ts
**Description**
The `typings/text` directory is lacking a blank `index.d.ts` file like all other directories in `typings` have. This leads to an error in vscode when using a `tsconfig.json` file in one's project that has a `typeRoots` pointing to the `moddable/typings` dir: "Cannot find type definition file for 'text'."

tve:
The typings/text directory does not have an index.d.ts file leading to errors coming from the language server in vscode (and presumably other editors). See https://github.com/Moddable-OpenSource/moddable/issues/1014. All other directories in typings have an index.d.ts.


phoddie:
Merged. Thank you.

PR #1026, opened 2023-02-04
example/io/tcp/mqtt: minor fixes

tve:
One commit fixes the options processing in mqtt.js
The other commit fixes the checking of the queue for keepalive pings in mqttclient.js

I'd like to also note that the keepalive algorithm in mqttclient.js is not according to the MQTT spec. Specifically, the connect message establishes a keepalive value. According to the 3.1.1 spec:

> The Keep Alive is [...] the maximum time interval that is permitted to elapse between the point at which the Client finishes transmitting one Control Packet and the point it starts sending the next. [MQTT-3.1.2-23].

Thus the time is about mqttclient _sending_ packets. However, the code in `#keepalive` looks at `this.#options.last` which is about when the last data was _received_ from the server. Thus there is a mismatch. For example, if the client does not publish anything but keeps receiving a stream of incoming messages then the algorithm will not send any pings and the server will disconnect it. It's not a big issue but it should be fixed at some point (IMO).

### DIFF
--- a/examples/io/tcp/mqtt/mqtt.js
+++ b/examples/io/tcp/mqtt/mqtt.js
@@ -69,7 +69,7 @@ export class Client {
 		let will = null;
 		if (options) {
 			if ("clientId" in options)
-				id = clientId
+				id = options.clientId
 			if ("keepalive" in options)
 				keepalive = 1000 * options.keepalive
 			if ("password" in options)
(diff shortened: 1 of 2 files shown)


phoddie:
Thanks for the fixes. We'll get those merged.

Indeed, the keepalive looks confused. It should be on writes by the client, not received data from the broker. I've made a note to fix that.

tve:
I think you need both: read and writes. Writes to keep the broker happy, and reads to notify the next layer up that something's amiss.

phoddie:
Closing, as the changes are merged.

PR #1030, opened 2023-02-11
FlashMode=DIO

NW-Lab:
Fixed to DIO because there is an individual that does not work in QIO mode with M5AtomS3


phoddie:
Thank you for the PR! I don't have an M5AtomS3 to test this. @kitazaki, would you check?

kitazaki:
@phoddie 
This works fine with my M5AtomS3 and macOS environment.

esptool.py esp32s3 -p /dev/cu.usbmodem14101 -b 921600 --before=default_reset --after=hard_reset write_flash --flash_mode dio --flash_freq 80m --flash_size 8MB 0x0 bootloader/bootloader.bin 0x10000 xs_esp32.bin 0x8000 partition_table/partition-table.bin

phoddie:
@kitazaki – thanks for the quick confirmation. I guess DIO is probably always safe, it just isn't always optimal as QIO is faster. But... if QIO is unreliable on this device, the default should be DIO. We'll get this merged.

PR #1033, opened 2023-02-14
Add typing for Console module

chigix:


phoddie:
Thanks for the PR. There are a couple of corrections:

- `receive` returns a number (`receive(): number`)
- `write` accepts one or more strings (`line(...items: string): void`)

Would you be willing to update the PR? If not, we can apply the changes after merging.

chigix:
@phoddie Thanks for you advice. I pushed new fix on to the Console type definition file.

chigix:
Additionally, is it possible to request more details in the usage example of Console class?
https://github.com/Moddable-OpenSource/moddable/blob/public/examples/base/console/main.js
In the existing example code, it seems the usage of Console class itself is not well presented, where why CLI would be used in the Console example seems also not comprehensible.

phoddie:
Thanks for the update. We've merged that.

The CLI Class is a way to build commands that are executed from some kind of terminal interface. The CLI is transport independent. The CLI is used in many places:
-  the base/console example with a simple serial transport
- the javascript/repl example with different transports (serial on microcontrollers, terminal on macOS, etc)
- the network/telnet example with the telnet protocol to allow remote commands
There is no requirement to use the CLI for these. You can rewrite them anyway you like. ;) The CLI class is just a convenient way to organize commands so that they can be re-used in different environments.

PR #1060, opened 2023-03-11
Extend wavstreamer request

meganetaaan:
This PR adds options to the WavStreamer class, allowing users to extend HTTP requests.

Use cases are provided in the Stack-chan repository.

https://github.com/meganetaaan/stack-chan/blob/cfc19242e6b13ebfa2ce1858bb9f5e96d1bc0c2d/firmware/stackchan/speech/tts-voicevox.ts#L87-L99

Some streaming servers accept requests other than GET (such as POST methods or requests for specific headers). By providing users with the ability to modify the request content, except for the properties required for the operation of WavStreamer, through callbacks such as onHeaders, this change enables WavStreamer to support such servers.

### DIFF
--- a/examples/pins/audioout/http-stream/readme.md
+++ b/examples/pins/audioout/http-stream/readme.md
@@ -73,6 +73,7 @@ The options object may contain the following properties. The `http`, `host`, `pa
 - `host` - the HTTP host to connect to stream from
 - `port` - the remote port to connect to
 - `path` - the path of the HTTP resource to stream
+- `request` - an optional object that can contain additional [HTTP request](../../../../documentation/network/network.md#http-request) options such as `method`, `headers`, `body` etc. These options will be passed directly to the HTTP client. If not specified, default options will be used.
 - `audio.out` - the audio output instance to play the audio on
 - `audio.stream` - the stream number of the audio output to use to play the audio. Defaults to `0`.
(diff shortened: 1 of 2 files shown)


phoddie:
@meganetaaan – This is a very reasonable request (and I like how you used ECMA-419 style to implement it so elegantly). Would you mind updating the readme as well, so other developers can more easily discover this capability?

FWIW – Allowing the caller to provide headers creates some complexity for future WavStreamer features. For example, to implement reconnect or seek to time, WavStreamer needs to send an HTTP range header. To do that, it will need to merge the caller's headers with its own. That's not a problem, just an implication.

meganetaaan:
I added descriptions for `request` option. Please check it.

>FWIW – Allowing the caller to provide headers creates some complexity for future WavStreamer features. For example, to implement reconnect or seek to time, WavStreamer needs to send an HTTP range header. To do that, it will need to merge the caller's headers with its own. That's not a problem, just an implication.

Got it. I understand that allowing the caller to provide headers can create complexity for future WavStreamer features. As long as we follow the rule of overriding essential elements as WavStreamer, and take care not to make the merging process too complicated, I believe we can handle this.

phoddie:
All good. Thank you. We'll get this merged.

phoddie:
This has been merged. Thank you!

PR #1067, opened 2023-03-20
xst: disable lsan when xs exit imminent

raphdev:
This PR affects only oss-fuzz build target for xst.

When ASAN is enabled, by default it reports all allocated memory at exit as memory leaks. However, this interferes with patterns that do not free memory when program exit is imminent as the allocated memory will be reclaimed at exit anyway. In XS, generally this is the case when fatal errors occur; those reports are not true memory leaks. 

Libfuzzer, as an in-process fuzzer, additionally tracks the ratio of `malloc` to `free` calls during each test case to determine if it should call LeakSanitizer's interface for on-demand leak detection. It uses the on-demand check to [find leaks during test cases](https://github.com/llvm/llvm-project/blob/release/15.x/compiler-rt/lib/fuzzer/FuzzerLoop.cpp#L673) instead of ASAN's default at exit detection. 

Ideally, we want leak detection to work, while not reporting leaks that would have been immediately freed at program exit. 

To do this, we implement [`__lsan_is_turned_off`](https://github.com/llvm/llvm-project/blob/main/compiler-rt/include/sanitizer/lsan_interface.h#L68) to disallow leak checking at certain parts of the program, independent of whether libfuzzer enables or disables lsan internally. The documentation states that it must return a constant value, but for our case it works as expected ([lsan check impl](https://github.com/llvm/llvm-project/blob/release/15.x/compiler-rt/lib/lsan/lsan_common.cpp#L710) for reference). With this change, we no longer need to suppress callsites where allocations may not be freed on fatal errors.

Tested against all current variants of false-positive memory leaks, and they no longer reproduce.

A tradeoff may be more out-of-memory issues reported, since libfuzzer is in-process and does not exit when XS would exit.

### DIFF
--- a/xs/tools/xst.c
+++ b/xs/tools/xst.c
@@ -237,7 +237,16 @@ static char *gxAbortStrings[] = {
 static txAgentCluster gxAgentCluster;
 
 #if OSSFUZZ
+static int lsan_disabled = 0;
+
+// allow toggling ASAN leak-checking
+__attribute__((used))
+int __lsan_is_turned_off() {
+	return lsan_disabled;
+}
+
 int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+	lsan_disabled = 0;
     fuzz_oss(Data, Size);
     return 0;
 }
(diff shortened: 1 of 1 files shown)


raphdev:
Closing as this got merged outside PR.

PR #1075, opened 2023-03-27
Adding SEN0322 O2 Sensor Driver Support

dabdoue:
Created sensor driver as well as example project

### DIFF
--- a/modules/drivers/sensors/sen0322/sen0322.js
+++ b/modules/drivers/sensors/sen0322/sen0322.js
@@ -1,0 +1,80 @@
+import Timer from "timer";
+
+const Register = Object.freeze({
+	OXYGEN_DATA: 0x03, ///< register for oxygen data
+	USER_SET: 0x08, ///< register for users to configure key value manually
+	AUTUAL_SET: 0x09, ///< register that automatically configure key value
+	GET_KEY: 0x0A ///< register for obtaining key value
+});
+
+const I2C_ADDR = 0x73;
+
+class SEN0322
+{
+	#io;
+	#key;
+	#onError;
+
+	constructor(options) {
+		this.#io = new options.sensor.io({
+			address: I2C_ADDR,
+			hz: 100_000,
+			...options.sensor
+		});
+		this.#onError = options.onError
+		
+	}
+
+	configure(options)
+	{
+		if (undefined !== options.vol)
+		{
+			const vol = options.vol;
+			const mv = options.mv ?? 0;
+			if (vol < 0 || vol > 25)
+				throw new RangeError("invalid O2 concentration (must be 0-25)");
+			else
+			{
+				const io = this.#io;
+				let keyValue = vol * 10;
+
+				if (mv < 0.000001 && mv > (-0.000001))		
+					io.write(Uint8Array.of(Register.USER_SET, keyValue));
+				else
+				{
+					keyValue = (vol / mv) * 1000;
+					io.write(Uint8Array.of(Register.AUTUAL_SET, keyValue));
+				}
+			}
+		}
+	}
+	#setKey()
+	{
+		const io = this.#io;
+		io.write(Uint8Array.of(Register.GET_KEY));
+		const curKey = (new Uint8Array(io.read(1)))[0];
+		if (curKey)
+			this.#key = curKey / 1000.0;
+		else
+			this.#key = 20.9 / 120.0;
+	}
+	close()
+	{
+		this.#io.close();
+		this.#io = undefined;
+	}
+	sample()
+	{
+		const io = this.#io;
+		let ret = {};
+		this.#setKey();
+		io.write(Uint8Array.of(Register.OXYGEN_DATA));
+		Timer.delay(100);
+		const rxbuf = new Uint8Array(io.read(3));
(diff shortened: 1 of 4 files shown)


modules/drivers/sensors/sen0322/sen0322.js
```diff
@@ -0,0 +1,80 @@
...
+	#key;
+	#onError;
+
+	constructor(options) {
+		this.#io = new options.sensor.io({
+			address: I2C_ADDR,
+			hz: 100_000,
+			...options.sensor
+		});
+		this.#onError = options.onError
```
phoddie: #onError is unused

modules/drivers/sensors/sen0322/sen0322.js
```diff
@@ -0,0 +1,80 @@
...
+	GET_KEY: 0x0A ///< register for obtaining key value
+});
+
+const I2C_ADDR = 0x73;
+
+class SEN0322
+{
+	#io;
+	#key;
+	#onError;
```
phoddie: #onError is unused

modules/drivers/sensors/sen0322/sen0322.js
```diff
@@ -0,0 +1,80 @@
...
+
+	configure(options)
+	{
+		if (undefined !== options.vol)
+		{
+			const vol = options.vol;
+			const mv = options.mv ?? 0;
+			if (vol < 0 || vol > 25)
+				throw new RangeError("invalid O2 concentration (must be 0-25)");
+			else
```
phoddie, line 62: The throw above will always, so the content of this else clause can be directly inline below.

modules/drivers/sensors/sen0322/sen0322.js
```diff
@@ -0,0 +1,80 @@
...
+		io.write(Uint8Array.of(Register.GET_KEY));
+		const curKey = (new Uint8Array(io.read(1)))[0];
+		if (curKey)
+			this.#key = curKey / 1000.0;
+		else
+			this.#key = 20.9 / 120.0;
+	}
+	close()
+	{
+		this.#io.close();
```
phoddie: close is defined to be [safe to call more than once](https://419.ecma-international.org/#-8-base-class-pattern-close-method). This implementation will throw after the first time it is called. We typically use `this.#io?.close()` here instead.

modules/drivers/sensors/sen0322/sen0322.js
```diff
@@ -0,0 +1,80 @@
...
+		this.#io = undefined;
+	}
+	sample()
+	{
+		const io = this.#io;
+		let ret = {};
```
phoddie: This is allocated very far from when it is used. Maybe simplify to:

```js
return {O: oxygenPercent * 10_000}; // O2 concentration in PPM
```

modules/drivers/sensors/sen0322/sen0322.js
```diff
@@ -0,0 +1,80 @@
+import Timer from "timer";
```
phoddie, line 24: Our ECMA-419 sensor drivers usually include a header with:

- Copyright notice
- Full name of part
- Link to relevant manufacturer data sheets
- Link to any implementations used as a reference

modules/drivers/sensors/sen0322/sen0322.js
```diff
@@ -0,0 +1,80 @@
...
+	}
+	sample()
+	{
+		const io = this.#io;
+		let ret = {};
+		this.#setKey();
+		io.write(Uint8Array.of(Register.OXYGEN_DATA));
+		Timer.delay(100);
+		const rxbuf = new Uint8Array(io.read(3));
+		const oxygenPercent = ((this.#key) * ((rxbuf[0]) + (rxbuf[1] / 10.0) + (rxbuf[2] / 100.0)));
```
phoddie: There's a diversity of approaches to parenthesis in this source file.

For example, this line is very careful about parenthesizing just about everything (including this.#key and the entire expression).

Compare to line 42 which only parenthesizes a number literal (`mv < 0.000001 && mv > (-0.000001)`). More consistent style with this repository would be (`(mv < 0.000001( && (mv > -0.000001))`) or a better way to express between, (`-0.000001 < mv) && (mv < 0.000001)`). That could apply to line 34 as well.

phoddie (changes requested):
Looks good. Here are some notes on changes to get it ready to merge.

modules/drivers/sensors/sen0322/sen0322.js
```diff
@@ -0,0 +1,96 @@
...
+	{
+		this.#io?.close();
+		this.#io = undefined;
+	}
+	sample()
+	{
+		const io = this.#io;
+		this.#setKey();
+		io.write(Uint8Array.of(Register.OXYGEN_DATA));
+		Timer.delay(100);
```
tve, line 89: Hmm, this pauses the VM. I would not use this driver for this alone. Would be nice to find a different way to code this. At the very least document it at the top or readme.

phoddie: Agreed. A simple solution would be to have a timer in the driver to poll the value every 100ms. The driver would store the most recent value and return it from `sample()`.

Because `sample()1 is allowed to return `undefined`, the driver could clear the cached sample after it is read once. The benefit is that a caller will know when they have a truly new reading.

PR #1085, opened 2023-04-12
Add X9C Potentiometers and TTP223 Switch Driver Support

dabdoue:
drivers and example projects included

### DIFF
--- a/modules/drivers/peripherals/x9c/x9c.js
+++ b/modules/drivers/peripherals/x9c/x9c.js
@@ -1,0 +1,172 @@
+/*
+ * Copyright (c) 2023  Moddable Tech, Inc.
+ *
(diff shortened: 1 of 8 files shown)


phoddie:
Looks great!

modules/drivers/peripherals/x9c/x9c.js
```diff
@@ -0,0 +1,172 @@
...
+                pin: csNVRAMWrite.pin
+            });
+        }
+        catch (e)
+        {
+            this.close();
+            throw e;
+        }
+
+        this.#reset(1);
```
tve, line 65: It would be great if one could pass in an initial value to avoid glitching to 1 if that's not the desired init value...

PR #1100, opened 2023-05-04
fix color inverted on M5Stack FIRE

washishi:
fix color inverted  on IPS LCD  (Import screen.js of m5stack)


phoddie:
@washishi - welcome and thank you for the PR!

Your changes look straightforward. I did not know that there is a variation of M5Stack Fire that uses an IPS screen. Did you test this change on both the original and IPS version of the hardware?

washishi:
No, I only have the IPS version, so I've only tried that one.

phoddie:
No problem. I have the original at my office. I will try that tomorrow.

phoddie:
Unfortunately, on the original M5Stack Fire the PR also inverts the colors.



So... maybe we have two choices:

- We need a way to identify the IPS display in software for M5Stack Fire
- We can give the user a way to indicate it is an IPS display when building for M5Stack Fire

How does the M5Stack Arduino software handle this?

washishi:
thank you for the test
Same as M5Stack
It should be possible to identify it by pulling up or pulling down GPIO33 before LCD initialization
I will check it a bit more

phoddie:
Interesting. I will get out the original M5Stack Fire again and give it a try. Fingers crossed!

washishi:
@phoddie
I tried internal pulldown on GPIO33 when investigating the LCD type
(In the IPS version, this is pulled up externally to identify it.)
Can you test again with the original TN LCD version?

The first change is referencing the screen.js used in targets/m5stack, so the TN LCD version of M5Stack BASIC or M5Stack Gray may have the same issue
I don't have a TN LCD version so I can't confirm...

phoddie:
Good news. You latest version works correctly with the original M5 Stack Fire! The value os `isIps` is false.

> The first change is referencing the screen.js used in targets/m5stack, so the TN LCD version of M5Stack BASIC or M5Stack Gray may have the same issue

I understand. Since the hardware is all the same, do you think we should modify m5stack/screen.js instead of creating a new version for M5Stack Fire only?

washishi:
@phoddie
> Good news. You latest version works correctly with the original M5 Stack Fire! The value os isIps is false.

Thank you for the operation test. I'm glad it seems to have worked properly.


> I understand. Since the hardware is all the same, do you think we should modify m5stack/screen.js instead of creating a new version for M5Stack Fire only?

Yes, I think you also need to fix screen.js in targets\m5stack
Not only M5Stack Fire, but also M5Stack BASIC and M5Stack Gray have switched to the IPS LCD version on the way, so the hardware around the LCD should be the same.

phoddie:
We agree. Would you please update this PR one last time?
- Apply the contents of m5stack_fire/screen.js to m5stack/screen.js
- Change the m5stack_fire manifest to use m5stack/screen.js (your original approach)

Also, if possible, please complete our [Contributor License Agreement](https://github.com/Moddable-OpenSource/moddable/blob/public/contributing.md#submitting-pull-requests).

Thank you!

washishi:
i fixed
IPS LCD versions of BASIC, Gray, and Fire have been confirmed to work with the balls demo.
I also filled in the Contributor License Agreement and sent it by email.
Thank you

phoddie:
Everything looks good! We will get this merged and live later this week. Thank you!

phoddie:
Closing - this has been merged.

PR #1108, opened 2023-05-12
Update default TLS protocol version to 1.2

jethrowharton:
I encountered an error while developing my app:
```
You are using TLS version 1.0, 1.1 and/or 3DES cipher which are deprecated to improve the security posture of Azure AD.
```
It seems that many services are switching their TLS versions to 1.2. It could make sense to have TLS 1.2 as the default protocol version.

I have tested this in my app, and it works well.

### DIFF
--- a/modules/crypt/ssl/ssl_constants.js
+++ b/modules/crypt/ssl/ssl_constants.js
@@ -62,7 +62,7 @@ export const NONE = 0;
 export const CBC = 1;
 export const GCM = 2;
 
-export const protocolVersion = (3 << 8) | 1;	// default protocol version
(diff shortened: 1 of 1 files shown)


phoddie:
@jethrowharton – Thanks for the PR. I think you are correct that it is time to update the default TLS version. We'll look at merging that.

By the way, from the error message it looks like you may be working with Azure Active Directory. I imagine that is of interest to other developers. If you have a chance to contribute a small example, kind of like @dabdoue did for [AWS](https://github.com/Moddable-OpenSource/moddable/tree/public/examples/network/aws) recently, that would be great. I

phoddie:
Closing as this was merged.

PR #1122, opened 2023-05-25
Add support for M5Stamp S3

stc1988:


phoddie:
Thank you for the PR for the new device support! It looks like maybe this the first port you've done for a USB-only device. Congratulations for working through that.

It looks like the documentation changes were merged on an earlier version -- the entire section on USB was deleted, for example. Would you check that?

stc1988:
Oops, I messed up the merge.
I fixed it.

phoddie:
Thanks for the quick update. It looks good now. We'll get this merged.

PR #1136, opened 2023-06-03
docs: add editorconfig for markdown

reconbot:
I was trying to fix a link in in a document and couldn't get github, my editor or my local git to preserve the mixed line endings in the file. Instead of doing that I figured I could help fix it.

[Editorconfig](https://editorconfig.org/#) is natively supported by most popular editors and keeps formatting issues like this away.

- First commit sets up editorconfig for markdown files using the most commonly used line endings.
- Second commit formats all the markdown files.


phoddie:
Thank you for the fix. We'll get that merged. It can only make life better.

(How embarrassing that we have inconsistent line endings)

PR #1144, opened 2023-06-08
Update modServo.c   Support ESP32S3

washishi:
ESP32S3 did not work because LEDC_HIGH_SPEED_MODE and LEDC_TIMER_15_BIT were undefined, so in that case I modified it to use LEDC_LOW_SPEED_MODE and LEDC_TIMER_14_BIT respectively.

### DIFF
--- a/modules/pins/servo/esp32/modServo.c
+++ b/modules/pins/servo/esp32/modServo.c
@@ -36,7 +36,11 @@ void xs_servo_destructor(void *data)
 	if (data) {
 		modServo ms = (modServo)data;
 		if (ms->running){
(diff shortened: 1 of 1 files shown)


washishi:
Confirmed that the servo, servo-rotation, and servo-sweep samples in examples/pins work properly on ESP32S3 (M5Stack CoreS3)
In addition, it may work properly with ESP32S2 and C series, but I have not verified it because I do not have the environment

phoddie:
@washishi – this looks great! Thank you for the work to sort this out and test. Very cool.

mkellner:
Committed https://github.com/Moddable-OpenSource/moddable/commit/b68cd76a81c3b1115812c85d5806f6d5016b548a

PR #1182, opened 2023-07-27
Fix typings for SNTP

meganetaaan:
I found a small mistake of a callback message of SNTP typings. Here is the fix.


phoddie:
Oops. We'll get that merged. Thank you!

phoddie:
This has been [merged](https://github.com/Moddable-OpenSource/moddable/commit/8e93df6c67b04f414893fc8636029efb3c95cf5d).

PR #1191, opened 2023-08-05
New module uploaded

salarizadi:
The cluster module was added to the modules/io folder
And its usage guide was added to the documentation/io folder


tve:
This looks interesting! Have you considered using the Worker interface instead and creating something like a RemoteWorker? (Maybe that already exists in JS land...)
Nitpick: I consider a "cluster" to be more than 2 machines... To me, the notion of cluster also includes some global management as opposed to a bunch of machines that happen to talk 1-1.

salarizadi:
> This looks interesting! Have you considered using the Worker interface instead and creating something like a RemoteWorker? (Maybe that already exists in JS land...) Nitpick: I consider a "cluster" to be more than 2 machines... To me, the notion of cluster also includes some global management as opposed to a bunch of machines that happen to talk 1-1.

This module can connect multiple esp32 boards or other expansion boards with UART, which increases the speed and efficiency so that you can transfer information easily. I think what you mean by RemoteWorker is that they connect to each other remotely, I don't quite understand.

tve:
What I meant by RemoteWorker is to abstract the connection and the concept of running something remotely using something patterned after the Worker class. Maybe that doesn't make sense, I haven't looked into it very deeply...

PR #1203, opened 2023-08-28
ECMA-419 BM8563 

stc1988:
This PR is a port of `modules/drivers/rtc/bm8563.js`.

In addition, I made small fixes for `m5stack_cores3`.
* add `bm8563` as built-in peripheral.
* remove startupVibration.


modules/drivers/peripherals/bm8563/bm8563.js
```diff
@@ -0,0 +1,113 @@
...
+  return v;
+}
+function bcdToDec(b) {
+  let v = (b / 16) | 0;
+  v *= 10;
+  v += b % 16;
+  return v;
+}
+
+Object.freeze(BM8563.prototype);
```
phoddie: Just FYI – this is unnecessary now because the XS linker automatically freezes all preloaded class prototypes. It was necessary in the past.

phoddie:
Cool, thank you. We'll look at getting this merged.

modules/drivers/peripherals/bm8563/bm8563.js
stc1988: Thank you for comment. I removed this.

phoddie:
Thanks! This has been merged and will be pushed soon.

PR #1215, opened 2023-09-22
examples/piu/outline/figures works on ESP32 devices

stc1988:
Some devices like `esp32/m5stack` needs to include `manifest_mod.json` to run mod.
This PR add include section to mods under `examples/piu/outline/figures`.


phoddie:
> Some devices like esp32/m5stack needs to include manifest_mod.json to run mod.

Yes, of course. We should always do that. Unfortunately, it is easy to overlook because it is not required for all devices. 

> This PR add include section to mods under examples/piu/outline/figures

Thank you. We'll get this merged.

PR #1229, opened 2023-10-07
Fix bleremotes link

stc1988:
I don't know if this is the only problem, I fixed broken link.


phoddie:
Thanks for the PR. If possible, we'd like the original link to work. We are checking to see if we can fix that. If not, we'll merge this.

phoddie:
It looks like the link incorrectly referred to an early draft of the blog post. The post was published at `blog/bleremotes/`  We'll merge this. Thank you again.

PR #1240, opened 2023-11-02
Compartment.prototype.@@toStringTag property

kriskowal:
I propose two changes.

1. Make `lockdown` available generally, but specifically so that test262 tests under `xst` can be framed for the language before and after `lockdown()`.
2. Adds a post-lockdown test for `Compartment.prototype.@@toStringTag` by way of demonstration
3. ~Consistent with my intuition, `Compartment.prototype.@@toStringTag` should be writable and configurable before `lockdown`. The pre-lockdown environment should be profanely mutable so that a shim can evolve the language, including the subsequent behavior of `lockdown`.~

### DIFF
--- a/tests/xs/built-ins/Compartment/prototype/Symbol.toStringTag-lockdown.js
+++ b/tests/xs/built-ins/Compartment/prototype/Symbol.toStringTag-lockdown.js
@@ -1,0 +1,13 @@
+/*---
+description:
+includes: [propertyHelper.js]
+features: [Symbol.toStringTag]
+---*/
+
+lockdown();
+
+assert.sameValue(Compartment.prototype[Symbol.toStringTag], 'Compartment');
+
+verifyNotEnumerable(Compartment.prototype, Symbol.toStringTag);
+verifyNotWritable(Compartment.prototype, Symbol.toStringTag);
+verifyNotConfigurable(Compartment.prototype, Symbol.toStringTag);
--- a/xs/tools/xst.c
+++ b/xs/tools/xst.c
@@ -350,6 +350,7 @@ int main(int argc, char* argv[])
 				xsSet(xsGlobal, xsID("gc"), xsResult);
 				xsResult = xsNewHostFunction(fx_fillBuffer, 2);
 				xsSet(xsGlobal, xsID("fillBuffer"), xsResult);
+#endif
 
 				xsResult = xsNewHostFunction(fx_harden, 1);
 				xsDefine(xsGlobal, xsID("harden"), xsResult, xsDontEnum);
@@ -359,7 +360,6 @@ int main(int argc, char* argv[])
 				xsDefine(xsGlobal, xsID("petrify"), xsResult, xsDontEnum);
 				xsResult = xsNewHostFunction(fx_mutabilities, 1);
 				xsDefine(xsGlobal, xsID("mutabilities"), xsResult, xsDontEnum);
-#endif
 
 				xsVar(0) = xsUndefined;
 				the->rejection = &xsVar(0);


kriskowal:
As written, this change would establish a precedent of suffixing `-lockdown` to post-lockdown test262 tests. I’m looking for a precedent but I’d welcome alternatives. It also strikes me that `lockdown` might need to be added to the `features` in front-matter, but I’m not sure what it would do.

tests/xs/built-ins/Compartment/prototype/Symbol.toStringTag-lockdown.js
```diff
@@ -0,0 +1,13 @@
...
+features: [Symbol.toStringTag]
+---*/
+
+lockdown();
+
+assert.sameValue(Compartment.prototype[Symbol.toStringTag], 'Compartment');
+
+verifyNotEnumerable(Compartment.prototype, Symbol.toStringTag);
+verifyNotWritable(Compartment.prototype, Symbol.toStringTag);
+verifyNotConfigurable(Compartment.prototype, Symbol.toStringTag);
```
gibson042, line 13:
```suggestion
verifyProperty(Compartment.prototype, Symbol.toStringTag, {
  value: 'Compartment',
  writable: false,
  enumerable: false,
  configurable: false
});
```

tests/xs/built-ins/Compartment/prototype/Symbol.toStringTag.js
```diff
@@ -7,5 +7,5 @@ features: [Symbol.toStringTag]
 assert.sameValue(Compartment.prototype[Symbol.toStringTag], 'Compartment');
 
 verifyNotEnumerable(Compartment.prototype, Symbol.toStringTag);
-verifyNotWritable(Compartment.prototype, Symbol.toStringTag);
+verifyWritable(Compartment.prototype, Symbol.toStringTag);
```
gibson042: Why should this change? The general pattern for prototype-object Symbol.toStringTag properties is to be configurable but not writable (as specified for e.g. [`Map.prototype`](https://tc39.es/ecma262/multipage/keyed-collections.html#sec-map.prototype-@@tostringtag)).
```console
$ eshost -se '[...
  Reflect.ownKeys(globalThis)
  .flatMap(key => {
    const proto = globalThis[key]?.prototype;
    if (typeof key !== "string" || !/^[A-Z][a-z]/.test(key) || !proto) return [];
    const { getOwnPropertyDescriptor: gOPD, getPrototypeOf: getProto } = Object;
    const tagDesc = (function getTag(obj){
      return obj ?
        gOPD(obj, Symbol.toStringTag) || getTag(getProto(obj)) :
        undefined;
    })(proto);
    if (!tagDesc) return [];
    return [[key, tagDesc]];
  })
  .reduce((m, [name, desc]) => {
    const { value: _value, get: _get, set: _set, ...meta } = desc;
    const schema = JSON.stringify(meta);
    const prevCount = m.get(schema);
    m.set(schema, (prevCount || 0) + 1);
    return m;
  }, new Map())
  .keys()
].join("\n")'
#### engine262, GraalJS, Hermes, JavaScriptCore, Moddable XS, QuickJS, SpiderMonkey, V8
{"writable":false,"enumerable":false,"configurable":true}
{"enumerable":false,"configurable":true}
```

xs/sources/xsModule.c
```diff
@@ -181,7 +181,7 @@ void fxBuildModule(txMachine* the)
 	slot = fxNextHostFunctionProperty(the, slot, mxCallback(fx_Compartment_prototype_import), 1, mxID(_import), XS_DONT_ENUM_FLAG);
 	slot = fxNextHostFunctionProperty(the, slot, mxCallback(fx_Compartment_prototype_importNow), 1, mxID(_importNow), XS_DONT_ENUM_FLAG);
 // 	slot = fxNextHostFunctionProperty(the, slot, mxCallback(fx_Compartment_prototype_module), 1, mxID(_module), XS_DONT_ENUM_FLAG);
-	slot = fxNextStringXProperty(the, slot, "Compartment", mxID(_Symbol_toStringTag), XS_DONT_ENUM_FLAG | XS_DONT_SET_FLAG);
+	slot = fxNextStringXProperty(the, slot, "Compartment", mxID(_Symbol_toStringTag), XS_DONT_ENUM_FLAG);
```
gibson042:
```suggestion
	slot = fxNextStringXProperty(the, slot, "Compartment", mxID(_Symbol_toStringTag), XS_DONT_ENUM_FLAG | XS_DONT_SET_FLAG);
```

gibson042 (commented):
> 3. Consistent with my intuition, `Compartment.prototype.@@toStringTag` should be writable and configurable before `lockdown`.

I would expect configurable but not writable (see specific comment below).

> As written, this change would establish a precedent of suffixing `-lockdown` to post-lockdown test262 tests. I’m looking for a precedent but I’d welcome alternatives.

That seems reasonable to me; it will place files that frequently need to be modified together adjacent in name-based sorting.

> It also strikes me that `lockdown` might need to be added to the `features` in front-matter, but I’m not sure what it would do.

It would allow for easy filtering to select or exclude tests that use lockdown in any particular run. :+1:

tests/xs/built-ins/Compartment/prototype/Symbol.toStringTag-lockdown.js
phoddie: This is just a question of style, correct? The style we use has been convenient when looking at test failures and we'd probably stick with that. For new tests you develop, please do whatever is most convenient, of course.

tests/xs/built-ins/Compartment/prototype/Symbol.toStringTag.js
phoddie: Agreed. Configurable will change after `lockdown()`. This test could check before and after.

phoddie:
> 1. Make `lockdown` available generally, but specifically so that test262 tests under `xst` can be framed for the language before and after `lockdown()`.

Agreed. I mistakenly thought that was already the case (it is available for fuzzing, but not test262).

tests/xs/built-ins/Compartment/prototype/Symbol.toStringTag-lockdown.js
kriskowal: Sadly, `verifyNonConfigurable` is destructive and uses `try/delete` internally, so there’s more substance than style here.

tests/xs/built-ins/Compartment/prototype/Symbol.toStringTag.js
kriskowal: Thanks. I was mistaken and I’ll remove this edit!

tests/xs/built-ins/Compartment/prototype/Symbol.toStringTag-lockdown.js
kriskowal: (But, this is orthogonal and I’ll respect @phoddie’s verdict on whether to include this suggestion here.)

phoddie:
This was manually merged with slight changes (see [here](https://github.com/Moddable-OpenSource/moddable/commit/9abdd0ad76d51f4bccd3a720ea7dec90756f3a30), [here](https://github.com/Moddable-OpenSource/moddable/commit/0e942afcc4c1f7b695ae6e1f0c08c14020c565d6), and [here](https://github.com/Moddable-OpenSource/moddable/commit/0d6fdf26d5f9c419d664f94409e5aa3d1bd9c3f6)). Closing PR.

PR #1250, opened 2023-11-18
INA233 power monitoring driver

tve:
This PR adds support for the Texas Instruments INA233 power monitoring chip. It is typically used in battery systems to monitor battery voltage, current as well as overall power consumption. A special feature of this chip is to accumulate power internally allowing accurate
energy consumption measurements with relatively infrequent polling.

This driver is intended to conform with ECMA-419 and only supports a simple polling interface which is sufficient to keep track of what the battery pack is doing. It does not support a number of the chip's features, such as threshold alarms.

### DIFF
--- a/examples/drivers/sensors/ina233/main.ts
+++ b/examples/drivers/sensors/ina233/main.ts
@@ -1,0 +1,33 @@
+// TI INA233 power monitor usage example
+// Copyright © 2023 by Thorsten von Eicken.
+
+import Timer from "timer"
+import Time from "time"
+import INA233, { ADC, POLARITY } from "embedded:sensor/Power-Management/INA233"
+
+const sensor = new INA233({
+  sensor: {
+    ...device.I2C.default,
+    io: device.io.SMBus,
+  },
+  onError: err => trace(`INA233: ${err}\n`),
+
+  shuntOhms: 0.003,
+  maxCurrent: 30, // max current in amps (default: 10A)
+  averaging: 64, // number of samples to average (1,4,16,64,...1024), default:1
+  vTime: ADC.TIME_588, // voltage ADC conversion time (ADC.TIME_*), default:1.1ms
+  aTime: ADC.TIME_588, // current ADC conversion time (ADC.TIME_*), default:1.1ms
+  polarity: POLARITY.BOTH, // polarity (POLARITY.*), default:POLARITY.BOTH
+})
+
+trace(`INA233: sampleIval=${sensor.sampleInterval}us maxIval=${sensor.maxInterval(600)}ms\n`)
+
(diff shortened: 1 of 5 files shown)


phoddie:
This looks pretty great. It will be our first sensor implemented in TypeScript. ;)

I noted a couple of small things to correct. I'll write those up when I have a little time.

phoddie:
- **onError**. The IO callbacks (`onReadable`, `onWritable`, `onError`) are are not supposed to be called directly from an API. The idea is to avoid calling a function that then calls you back during its execution, as that is error prone (see "callbacks may not be invoked by the instance from within its public method calls, including the constructor" in the spec). Since this driver is fully synchronous, it doesn't need to use `onError`. 
- The call to `readBuffer` on line 105 will throw if the device isn't present. In that case, the #io will be orphaned.
- The constructor is potentially modifying the input options. It should make a copy before doing that. (See "The constructor must not modify the options object. It must accept an immutable options object" in the spec)
- But... the constructor is for setting up the hardware. Any changes to the default configuration should be applied by calling `configure`.

Combining these comments, you might rework the constructor:

```js
  constructor(options: Options) {
	try {
		const io = (this.#io = new options.sensor.io({
		  hz: 400_000,
		  address: 0x40,
		  ...options.sensor,
		}))

		// verify that we're talking to an ina233
		const modelBuf = this.#io.readBuffer(COMMANDS.MFR_MODEL, 7)
		const model = new TextDecoder().decode(new Uint8Array(modelBuf, 1))
		if (model !== "INA233")
			throw new Error(`Wrong model: ${model}`)

		this.#io.sendByte(COMMANDS.RESTORE_DEFAULT_ALL)

		// ensure certain configuration options get set now
		this.configure({shuntOhms: 0.01, maxCurrent: 10, averaging: 1});
	}
	catch (e) {
		this.close();
		throw e;
	}
  }
```

- **get sampleInterval**: This is defined to be in microseconds. ECMA-419 uses milliseconds for time everywhere it possibly can. This is the default in JavaScript. Being consistent in the APIs avoids confusion about units.

tve:
Thanks for the review! I think I need an "ECMA-419 driver checklist" 😆

phoddie:
> Thanks for the review! I think I need an "ECMA-419 driver checklist" 😆

Agenda item for the December 6 workshop: "Making ECMA-419 more approachable for developers". This is exactly the kind of thing that we should do. (cc @hipsterbrown @dtex).

tve:
Fixing this driver may take a little while 'cause of a testing bottleneck on my end. I only have one INA233 chip operational (dead-bugged) and I'm not sure I want to mess-up the firmware of the device it's in... I should make a breakout board...

PR #1251, opened 2023-11-18
Add new driver for HX711 load cell ADC

tve:
The moddable SDK already has a driver for the HX711 but I found it to be very, very slow. E.g. over 300ms per reading, in part due to averaging that can't be disabled. This new driver is very simple and optimized for speed: it essentially just provides one `read()` function to read the current ADC value without performing any averaging. There is a second function `readable()` to check that the chip can be read (it typically samples at 10Hz). This driver implements the read in C using modGPIO and performs a read in a couple of milliseconds.

Conveniently the existing driver is under `modules/drivers` and this PR places code in `modules/drivers/sensors` so there is no clash.

This driver is intended to conform to ECMA-419.

### DIFF
--- a/modules/drivers/sensors/hx711/hx711.ts
+++ b/modules/drivers/sensors/hx711/hx711.ts
@@ -1,0 +1,76 @@
+// Avia HX711 24-bit ADC for weight scales
+// Copyright © 2023 by Thorsten von Eicken.
+
+import Time from "time"
+import HX711c from "embedded:sensor/ADC/HX711c"
+
+export interface Options {
+  sensor: {
+    clk: number
+    din: number
+  }
+  onError?: (error: string) => void
+
+  // gain selects the gain of the ADC as well as the analog input channel
+  gain?: number // 1:128chA, 2:32chB, 3:64chA (default: 1)
+}
+
+export default class HX711 {
+  #hx711c: HX711c
+
+  constructor(options) {
+    this.configure(options)
+  }
+
+  close() {
+    this.#hx711c = undefined
+  }
+
+  // To "configure" the HX711 we have to perform a read, which ends up setting up the gain
+  // for the next read
+  configure(options: Options) {
+    this.#hx711c = new HX711c(options.sensor.clk, options.sensor.din, options.gain || 1)
+    this.#hx711c.read()
+  }
+
+  get format() {
+    return "number"
+  }
+
+  readable() {
+    //return this.#din.read() == 0
+  }
+
+  read() {
+    return this.#hx711c.read()
+  }
+
+  /* read() has to be implemented in C to meet the timing requirements (clk high < 50us)
+
+  // read the 24-bit signed value from the sensor and perform additional pulses to set-up the gain
+  read() {
+    const clk_w = this.#clk.write.bind(this.#clk)
+    const din_r = this.#din.read.bind(this.#din)
+    if (din_r() != 0) return undefined
+    // read 24 bits: din is stable 100ns after clk rising edge until next rising edge, so we read
+    // after producing the falling edge
+    // ugh: clk high must be less than 50us or the chip enters power-down mode
+    let val = 0
+    clk_w(0) // preload caches...
+    for (let i = 24; i > 0; i--) {
+      clk_w(1)
+      clk_w(0)
+      val = (val << 1) | (din_r() & 1)
+    }
+    // sign extend
+    val = (val << 8) >> 8
+    if (val == -1) return undefined // chip entered power-down mode
+    // pulse the clock to set the gain
+    for (let i = this.#gain; i > 0; i--) {
+      clk_w(1)
+      clk_w(0)
+    }
+    return val
+  }
+*/
+}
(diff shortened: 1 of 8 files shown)


modules/drivers/sensors/hx711/hx711.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  close() {
+    this.#hx711c = undefined
+  }
+
+  // To "configure" the HX711 we have to perform a read, which ends up setting up the gain
+  // for the next read
+  configure(options: Options) {
+    this.#hx711c = new HX711c(options.sensor.clk, options.sensor.din, options.gain || 1)
+    this.#hx711c.read()
```
phoddie, line 33: configure() can be called more than once. If it is, the first HX711c instance will be replaced with another. But, there's also no guarantee that `options.sensor` will be present when `configure` is called, since that property is for the constructor. The simplest solution is to move these two lines to the constructor and leave `configure` empty.

modules/drivers/sensors/hx711/hx711.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  get format() {
+    return "number"
+  }
+
+  readable() {
+    //return this.#din.read() == 0
+  }
```
phoddie, line 42: Is this function necessary? It isn't in the standard. The usual behavior is to have `sample()` return `undefined` if no reading is available. (It looks like this is what you implemented in your native `read` function already)

modules/drivers/sensors/hx711/modHX711c.c
```diff
@@ -0,0 +1,76 @@
...
+    xsUnknownError("can't init clk pin");
+  modGPIOWrite(&data->clk, 0);
+  if (modGPIOInit(&data->din, NULL, din_pin, kModGPIOInput))
+    xsUnknownError("can't init dat pin");
+
+  xsmcSetHostData(xsThis, data);
+}
+
+void xs_HX711_destructor(void *hostData) {
+  hx711_data *data = hostData;
```
phoddie, line 34: The destructor can be called with `hostData` set to `NULL`. This happens when the virtual machine is deleted. It is rare. ;) It can also happen if an exception occurs in the constructor before the host data is set. So, the best practice is to always check for `NULL`

modules/drivers/sensors/hx711/modHX711c.c
```diff
@@ -0,0 +1,76 @@
+// Copyright © 2023 by Thorsten von Eicken.
+
+#include "xsPlatform.h"
+#include "xsmc.h"
+#include "modGPIO.h"
+#include "mc.xs.h" // for xsID_* constants
+
+#define xsmcVar(x) xsVar(x)
```
phoddie, line 8: unused. what was the idea?

modules/drivers/sensors/hx711/modHX711c.c
```diff
@@ -0,0 +1,76 @@
...
+  int din_pin = xsmcToInteger(xsArg(1));
+
+  hx711_data *data = c_malloc(sizeof(hx711_data));
+  if (data == NULL) xsUnknownError("can't allocate data");
+  data->gain = xsmcToInteger(xsArg(2));
+  if (modGPIOInit(&data->clk, NULL, clk_pin, kModGPIOOutput))
+    xsUnknownError("can't init clk pin");
+  modGPIOWrite(&data->clk, 0);
+  if (modGPIOInit(&data->din, NULL, din_pin, kModGPIOInput))
+    xsUnknownError("can't init dat pin");
```
phoddie, line 28: If any of these throw, `data` is orphaned (even the call to `xsmcToInteger` can call in type coercion cases). Error clean-up isn't pretty, but here's what I'd probably do here:

```c
  int clk_pin = xsmcToInteger(xsArg(0));
  int din_pin = xsmcToInteger(xsArg(1));
  hx711_data d; 
  d.gain = xsmcToInteger(xsArg(2));

  if (modGPIOInit(&d.clk, NULL, clk_pin, kModGPIOOutput))
    xsUnknownError("can't init clk pin");
  if (modGPIOInit(&d.din, NULL, din_pin, kModGPIOOutput)) {
    modGPIOUninit(&d.clk);
    xsUnknownError("can't init din pin");
  }
  hx711_data *data = c_malloc(sizeof(hx711_data));
  if (data == NULL) {
    modGPIOUninit(&d.clk);
    modGPIOUninit(&d.din);
    xsUnknownError("can't allocate data");
  }
  data = d;
```

modules/drivers/sensors/hx711/modHX711c.c
```diff
@@ -0,0 +1,76 @@
...
+#define xsmcVar(x) xsVar(x)
+
+typedef struct {
+  modGPIOConfigurationRecord clk;
+  modGPIOConfigurationRecord din;
+  int gain;
+} hx711_data;
+
+void xs_HX711_init(xsMachine *the) {
+  if (xsmcArgc != 3) xsUnknownError("invalid arguments");
```
phoddie, line 17: This check isn't strictly necessary. Using xsArg on an argument that doesn't exist will throw. All this guarantees is that there aren't more than 3 arguments. No harm in the check, but it is common for JavaScript functions to simply ignore extra arguments.

PR #1253, opened 2023-11-18
typings for embedded:network (io) modules

tve:
This PR includes a number of typings for various network modules. It also includes some missing manifests with content copied from the associated examples. It's basically a list of fixes I needed to get my apps to build.

### DIFF
--- a/typings/embedded_network/dns/resolver/udp.d.ts
+++ b/typings/embedded_network/dns/resolver/udp.d.ts
@@ -28,8 +28,8 @@ declare module "embedded:network/dns/resolver/udp" {
 
   class Resolver {
     constructor(options: Record<string, any>)
-    close(): null
-    resolve(options: Options): null
+    close(): void
+    resolve(options: Options): void
 
   }
 
--- a/typings/embedded_network/dns/resolver/udp.d.ts
+++ b/typings/embedded_network/dns/resolver/udp.d.ts
@@ -1,0 +1,37 @@
+/*
+* Copyright (c) 2022 Shinya Ishikawa
+*
+*   This file is part of the Moddable SDK Tools.
+*
+*   The Moddable SDK Tools is free software: you can redistribute it and/or modify
+*   it under the terms of the GNU General Public License as published by
+*   the Free Software Foundation, either version 3 of the License, or
+*   (at your option) any later version.
+*
+*   The Moddable SDK Tools is distributed in the hope that it will be useful,
+*   but WITHOUT ANY WARRANTY; without even the implied warranty of
+*   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*   GNU General Public License for more details.
+*
+*   You should have received a copy of the GNU General Public License
+*   along with the Moddable SDK Tools.  If not, see <http://www.gnu.org/licenses/>.
+*
+*/
+
+declare module "embedded:network/dns/resolver/udp" {
+
+  interface Options {
+    host: string
+    onResolved: ((host:string, address:any) => null)
+    onError: ((host:string) => null)
+  }
+
+  class Resolver {
+    constructor(options: Record<string, any>)
+    close(): null
+    resolve(options: Options): null
+
+  }
+
+  export default Resolver
+}
--- a/examples/io/tcp/httpsclient/config.js
+++ b/examples/io/tcp/httpsclient/config.js
@@ -5,7 +5,7 @@ import UDP from "embedded:io/socket/udp";
 import Resolver from "embedded:network/dns/resolver/udp";
 
 import HTTPClient from "embedded:network/http/client";
-import TLSSocket from "tlssocket";
+import TLSSocket from "embedded:io/socket/tlssocket";
 
 const dns = {
 	io: Resolver,
--- a/examples/io/tcp/httpsclient/main.js
+++ b/examples/io/tcp/httpsclient/main.js
@@ -13,7 +13,7 @@
  */
 
 import TextDecoder from "text/decoder"
-import TLSSocket from "tlssocket";
+import TLSSocket from "embedded:io/socket/tlssocket";
 
 const http = new device.network.http.io({ 
 	...device.network.http,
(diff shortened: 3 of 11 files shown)


typings/embedded_network/dns/resolver/udp.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  interface Options {
+    host: string
+    onResolved: ((host:string, address:any) => null)
+    onError: ((host:string) => null)
+  }
+
+  class Resolver {
+    constructor(options: Record<string, any>)
+    close(): null
+    resolve(options: Options): null
```
phoddie: A return value of `null` is unusual. It seems like the usual thing in these cases is `void`?

phoddie:
This looks really good -- just a couple of minor points above to deal with.

typings/embedded_network/dns/resolver/udp.d.ts
tve: Thanks for catching this!

tve:
Issue corrected, thanks!

phoddie:
I'd like to get this merged, but I'm not quite sure of the status. It looks like one of the review issue was corrected, but GitHub still shows others as unresolved.

tve:
I believe all that was missing was me hitting the "resolved" button on github. I guess the review interface has changed since I last used it a bunch of years ago... LMK if there is anything else amiss.

phoddie:
Oh, OK. Thanks. Glad I asked. Let's see if we can get this one merged then.

phoddie:
Two points weren't addressed: using the ECMA-419 module specifier for TLS  and not using "embedded:" for the non-standard MQTT.js module. I've resolved those locally on the merged version. That'll get propagated to this repository later this week.

phoddie:
I have a kind of rookie question. ;)

After applying all this, I wanted to try using TypeScript on some of the ECMA-419 network examples. I stared with `$MODDABLE/examples/io/tcp/httpclient` and added manifest_typings to the project's manifest.json and renamed "main.js" to "main.ts". That gets me in trouble here:

```
main.ts:17:27 - error TS2339: Property 'network' does not exist on type '{ io: { ....
const HTTPClient = device.network.http.io;
```

The `$MODDABLE/typings/embedded_provider/builtin.d.ts` doesn't define `device.network`, so the error makes sense. I tried to add `device.network.http.io` but not successfully.

My impression is that both @tve and @meganetaaan are able to use TypeScript with httpclient. But, how?

tve:
I added network into the device type, specifically, in typings/embedded_provide/builtin.d.ts:
```
    SMBus: { default: ConstructorParameters<typeof SMBus>[0] };
    SPI: { default: ConstructorParameters<typeof SPI>[0] };
    Serial: { default: ConstructorParameters<typeof Serial>[0] };
    pin: Record<string, PinSpecifier>;                                                     
    network?: Record<string, any>;                                                         
  };
  export default device;
}
```
This is obviously not a clean solution, but got me through that hurdle. It's not entirely clear to me how strict that network section should be managed/typed.

I also have the following at the start of my main.ts:
```
declare global {
  var mcsim: boolean
}

// ensure global.device.network exists and is not frozen so we can add stuff to it as the network
// initializes -- note that in mcsim there is no globalThis.device
if ("device" in globalThis) {
  const dn = globalThis.device.network
  if (!dn) globalThis.device = { ...globalThis.device, network: {} }
  else if (Object.isFrozen(dn)) globalThis.device = { ...globalThis.device, network: { ...dn } }
  globalThis.mcsim = false
} else {
  globalThis.mcsim = true
}
```

phoddie:
> It's not entirely clear to me how strict that network section should be managed/typed.

The `network` section is normatively defined, so what is expected is defined.

Focusing on the httpcoient example, I added an import of HTTPClient to builtin.d.ts:

```js
  import HTTPClient from "embedded:network/http/client";
```
And then defined `network` like this (ignoring that `network` and `http` are optional for the moment)::

```js
    network: {
      http: {
        io: typeof HTTPClient;
      }
    }
```

That allows the following to work in main.ts:

```js
const HTTPClient = device.network.http.io;
const http = new HTTPClient({ 
```

It doesn't like the options object. TypeScript doesn't know that some of of the required properties come from `...device.network.http` and so reports an error. That can be quieted by declaring `device.network` with the required properties of the options object with type `any`:


```js
    network: {
      http: {
        io: typeof HTTPClient;
        socket: any;
        dns: any;
      }
    }
```

Whenever `any` appears in TypeScript, it seems like a compromise. But, onward.

TypeScript really hates this:

```js
			this.decoder = new TextDecoder;
```

It reports `error TS2339: Property 'decoder' does not exist on type 'HTTPRequest'.` Adding the following to `main.ts` seems to solve that by locally extending the http client interface.

```ts
declare module "embedded:network/http/client" {
	interface HTTPRequest {
		decoder: TextDecoder
	}
}
```

There are a couple of small errors in http/client.d.ts that cause problems with this example. The `onError` callback should be optional:

```ts
    onError?: (err:any)=>void
```

And the `onHeaders` callback should use a `Map` for the headers object:

```
    onHeaders?: (this: HTTPRequest, status:number, headers: Map<String, string>) => void
```

Then... it is possible to build the httpclient/main.ts example with TypeScript. The ugliest change is to support `this.decoder`. But that's in the app, not the ECMA-419 objects. The app could be written differently. Are the other changes reasonable though?

tve:
Makes sense. I had to change the opening lines of `main.ts` to:
```
const HTTPClient = device.network!.http!.io;
const http = new HTTPClient({ 
	...device.network!.http!,
	host: "example.com"
});
```
'cause I have `network` and `http` as optional. For grins, this is how copilot auto-completed the network section of `builtin.d.ts`:
```
    network?: {
      http?: {
        io: typeof import("embedded:network/http/client").default;
        dns: {
          io: typeof import("embedded:network/dns/resolver/udp").default;
          servers: string[];
          socket: {
            io: typeof import("embedded:io/socket/udp").default;
          };
        };
        socket: {
          io: typeof import("embedded:io/socket/tcp").default;
        };
      };
    }
```

What is the benefit or rationale for writing `const HTTPClient = device.network.http.io` instead of `import HTTPClient from "embedded:network/http/client"`?

phoddie:
It looks like copilot did a respectable job lifting from httpclient/config.js.

> What is the benefit or rationale for writing const HTTPClient = device.network.http.io instead of import HTTPClient from "embedded:network/http/client"?

The stuff hanging off `device` is from the host. Everything should be configured optimally for that host. If you are going to use `device.network.http` as the basis for the constructor options object you should use the `io` constructor it provides, since they are intended to be used together.

PR #1282, opened 2023-12-30
LTR553ALS  Light & Proximity Sensor

stc1988:


phoddie:
This looks really good. Thank you! I noticed a few things that should be corrected or could be improved. I'll do a quick code review. Please update as you see appropriate. Then we can merge this.

examples/drivers/sensors/ltr553als/main.js
```diff
@@ -0,0 +1,26 @@
...
+ *   or send a letter to Creative Commons, PO Box 1866,
+ *   Mountain View, CA 94042, USA.
+ *
+ */
+
+import Proximity from "embedded:sensor/Light-Proximity/LTR553ALS";
+import Timer from "timer";
+
+const sensor = new Proximity({
+  sensor: { ...device.I2C.internal, io: device.io.SMBus },
```
phoddie: Not all hosts have `device.I2C.internal` so may do `...(device.I2C.internal ?? device.I2C.default),` so it works on more devices.

modules/drivers/sensors/ltr553als/ltr553als.js
```diff
@@ -0,0 +1,342 @@
...
+  sample() {
+    const ps = this.psValue;
+    const alsCh1 = this.alsCh1Value;
+    const alsCh0 = this.alsCh0Value;
+    const als = (alsCh0 + alsCh1) >> 1;
+
+    return {
+      ps,
+      als,
+    };
```
phoddie: This is a compound sensor. That means it has both [AmbientLight](https://419.ecma-international.org/#-sensor-classes-ambient-light) and [Proximity](https://419.ecma-international.org/#-14-sensor-classes-proximity) sensors. The form of the returned object should be:

```js
{
   proximity: {
      distance: 123
   },
   lightmeter: {
      illuminance: 234
   }
}
```

modules/drivers/sensors/ltr553als/ltr553als.js
```diff
@@ -0,0 +1,342 @@
...
+
+    this.#setPsMode(PsMode.PS_ACTIVE_MODE);
+    this.#setAlsMode(AlsMode.ALS_ACTIVE_MODE);
+  }
+  close() {
+    this.#io.close();
+    this.#io = undefined;
+  }
+
+  get psValue() {
```
phoddie: These three accessors (getters) are unintentionally part of the public API of the sensor. I think they are just helper functions. Since they are only called in one place, maybe it is better to inline them into `sample()`. If you do that, you can also reuse `buffer` for all three reads, which will be a little lighter.

modules/drivers/sensors/ltr553als/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+	"modules": {
+		"embedded:sensor/Light-Proximity/LTR553ALS": "$(MODDABLE)/modules/drivers/sensors/ltr553als/ltr553als"
```
phoddie: `embedded:sensor/AmbientLight-Proximity/LTR553ALS`

modules/drivers/sensors/ltr553als/ltr553als.js
```diff
@@ -0,0 +1,342 @@
...
+    if (undefined != options.psMeasRate) {
+      let value = io.readUint8(Register.ALS_MEAS_PATE_REG);
+      io.writeUint8(Register.ALS_MEAS_PATE_REG, value & PS_MEAS_RATE_MASK);
+    }
+
+    this.#setPsMode(PsMode.PS_ACTIVE_MODE);
+    this.#setAlsMode(AlsMode.ALS_ACTIVE_MODE);
+  }
+  close() {
+    this.#io.close();
```
phoddie: `close` is defined to be [safe to call more than once](https://419.ecma-international.org/#-8-base-class-pattern-close-method). So, this should be `this.#io.close?.()`

modules/drivers/sensors/ltr553als/ltr553als.js
```diff
@@ -0,0 +1,342 @@
...
+      psLedDutyCycle: PsLedCurrentDuty.CURRENT_DUTY_PER100,
+      psMeasRate: PsMeasurementRate.PS_MEASUREMENT_RATE_50MS,
+      psLedPeakCurrent: PsLedPeakCurrent.LED_PEAK_CURRENT_100MA,
+      psNPulses: 1,
+      alsGain: AlsGain.ALS_GAIN_48X,
+      alsIntegrationTime: AlsIntegrationTiime.ALS_INTEGRATION_TIME_100MS,
+      alsMeasureRate: AlsMeasureRate.ALS_MEASUREMENT_RATE_500MS,
+    });
+
+    this.#onError = options.onError;
```
phoddie: This is unused. I think you can eliminate `#onError` completely.

stc1988:
Thank you for reviewing. I fixed some codes to your comment.
And I will check [ECMA-419 specification](https://419.ecma-international.org/).

phoddie:
The changes look good. Thanks for the quick work.

Since you are using the i2c.internal bus, I'm guessing that this driver is for a built-in sensor on a development board. Which board is it for?

stc1988:
I use it for M5Stack CoreS3.

https://docs.m5stack.com/en/core/CoreS3

phoddie:
Thanks for letting me know. I should have guessed that. ;)

phoddie:
(Merged - will be live soon)

PR #1286, opened 2024-01-05
OpenAI text-to-speech

stc1988:
Text-to-speech streaming API for the [OpenAI Create Speech](https://platform.openai.com/docs/api-reference/audio/createSpeech) like [elevenlabs-stream](https://github.com/Moddable-OpenSource/moddable/tree/public/examples/pins/audioout/elevenlabs-stream).

Demo
https://x.com/stc1988/status/1743308674406437345?s=20

### DIFF
--- a/examples/pins/audioout/openai-stream/main.js
+++ b/examples/pins/audioout/openai-stream/main.js
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2024  Moddable Tech, Inc.
+ *
+ *   This file is part of the Moddable SDK.
+ * 
+ *   This work is licensed under the
+ *       Creative Commons Attribution 4.0 International License.
+ *   To view a copy of this license, visit
+ *       <http://creativecommons.org/licenses/by/4.0>.
+ *   or send a letter to Creative Commons, PO Box 1866,
+ *   Mountain View, CA 94042, USA.
+ *
+ */
+
+import AudioOut from "pins/audioout"
+import OpenAIStreamer from "openaistreamer";
+
+const key = "PUT-YOUR-API-KEY-HERE";
+if (key.includes("API-KEY"))
+	throw new Error("Set your OpenAI API key");		// https://platform.openai.com/docs/guides/production-best-practices/api-keys
+
+const audio = new AudioOut({});
+audio.start();
+
+new OpenAIStreamer({
+	key,
+	model: "tts-1",
+	voice: "alloy",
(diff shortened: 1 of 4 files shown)


phoddie:
This looks really great! Very cool that you were able to re-use the basic pattern of the ElevenLabs TTS streamer. 

The demo sounds great. Is there anything we can do to reduce the delay at the start? I checked the OpenAI documentation but didn't see any options. I suppose there are two major parts of the delay -- establishing the TLS connection and waiting for the audio to begin to be delivered.

stc1988:
I added trace logs  and took PROFILE.　Can this help identify factors?

``` javascript
trace(`${Date()} OpenAIStreamer start\n`)
new OpenAIStreamer({
	key,
	model: "tts-1",
	voice: "alloy",
	input: "The Moddable SDK is a combination of development tools and runtime software to create applications for microcontrollers.",
	audio: {
		out: audio,
		stream: 0
	},
	onReady(state) {
		trace(`${Date()} OpenAI Ready: ${state}\n`)
	},
...
```
スクリーンショット 2024-01-06 8 53 35

phoddie:
Great idea to use the profiler.

I assume you started the profile before instantiating `OpenAIStreamer`. When did you stop it?

The biggest item is the garbage collector `(gc)`. That's not so surprising since MP3 streaming involves a lot of buffer operations, as does TLS. You can probably reduce that gc overhead by making the heap bigger (maybe increase `status` from to 110 KB). The GC overhead should be less because it will run less often.

Most of the rest of what is taking time is the TLS overhead, but it doesn't look all that unusual. You get from instantiating OpenAIStreamer to "ready" in about 3 seconds. Establishing the TLS connection uses the heaviest crypto operations. The delay in starting up in your video is much longer than 3 seconds, so TLS and GC are not the main reason.

When the video starts, is the device first connecting to Wi-Fi. Maybe that is taking some time?

(I'm interested in minimizing the latency so that Stack-chan can talk as soon as possible.)

stc1988:
> When the video starts, is the device first connecting to Wi-Fi. Maybe that is taking some time?

Yes, the video I took starts from launching microcontroller.

stc1988:
> I assume you started the profile before instantiating OpenAIStreamer. When did you stop it?

I stopped it while playback speech.

phoddie:
It seems like there's nothing very unusual here with performance. I think the main thing is that it wasn't clear to me that the video started with the microcontroller starting up and connecting to Wi-Fi. We'll get this merged.

PR #1289, opened 2024-01-12
JSON parser tests

markwharton:
This PR adds a time zone offset to ensure that tests pass in any region and fixes a broken link to the Moddable testing documentation.


phoddie:
Thank for the fix. We've merged this internally and will push public shortly.

PR #1292, opened 2024-01-17
ADD aw3641 driver

NW-Lab:
Hello

I made a driver for FLASH LAMP aw3641.
This is not ECMA-419 compliant. I could not find the item.
The product to use is M5Stack's FlashLamp.
https://docs.m5stack.com/en/unit/FlashLight

Thanks,

### DIFF
--- a/build/devices/esp32/targets/m5stack_core2/manifest.json
+++ b/build/devices/esp32/targets/m5stack_core2/manifest.json
@@ -12,7 +12,9 @@
 	],
 	"config": {
 		"screen": "ili9341",
-		"touch": "M5StackCoreTouch"
+		"touch": "M5StackCoreTouch",
+		"startupSound": "bflatmajor.maud",
+		"startupVibration": 600
 	},
 	"defines": {
 		"i2c": {
--- a/build/devices/esp32/targets/m5stack_core2/manifest.json
+++ b/build/devices/esp32/targets/m5stack_core2/manifest.json
@@ -12,9 +12,7 @@
 	],
 	"config": {
 		"screen": "ili9341",
-		"touch": "M5StackCoreTouch",
-		"startupSound": "bflatmajor.maud",
-		"startupVibration": 600
+		"touch": "M5StackCoreTouch"
 	},
 	"defines": {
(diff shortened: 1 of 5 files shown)


phoddie:
Thank you for the PR!

> This is not ECMA-419 compliant. I could not find the item

Thank you for considering this. While there is no ECMA-419 specification yet for the aw3641,  it could be implemented by applying the Peripheral Class Pattern as a starting point, and then adding an API to set the lamp level. Maybe even `write()` could work for that. (It is not required to do this; I'm just explaining how to apply ECMA-419 style for a new device.)

The one concern I have with the implementation is the long synchronous delay (over 1 second in some cases). That will block the entire device, which isn't ideal for JavaScript. I sketched out a small modification to the implementation to use a timer instead of a delay. What do you think of this idea?

```js
class AW3641 extends Digital {
    #timer;

	constructor(options = {}) {
		super({
			pin: options.pin,
            mode: Digital.Output
		});
	}
    flash(value){
        if (this.#timer) {
                this.#timer.next = value;	// wait until previous setting is complete
                return;
        }

        for(let i=0;i<value;i++){
            super.write(0);
            super.write(1);
        }
	
       let delay = 0;
        if((value>0)&&(value<9)) delay = 220;
        if((value>8)&&(value<17)) delay = 1300;
        this.#timer = Timer.set(() => {
			super.write(0);

			const next = this.#timer.next;
			this.#timer = undefined;
			if (undefined !== next)
				this.flash(next);
        }, delay);
   }
}
```

NW-Lab:
Hello,

I created a driver for the first time.
thank you for the advice.
We have confirmed that the modified code works properly. I think this is better.

build/devices/esp32/targets/m5stack_core2/manifest.json doesn't need to be changed, so I reverted it back.
I modified the example code a little.

thank you

phoddie:
> I created a driver for the first time.

That's great. Congratulations. 

> We have confirmed that the modified code works properly. I think this is better.

Thank you for giving that a try. Glad it worked. We will get this merged.

NW-Lab:
thank you

Issue #1236, opened 2023-10-23
No AudioOut sounds with M5Stack CoreS3


meganetaaan:
**Build environment:** Linux
**Moddable SDK version:** v4.2.0, v4.2.1
**Target device:** M5Stack CoreS3

**Description**
With M5Stack CoreS3 and the ModdableSDK v4.2.1, no sound can be played through the AudioOut class, leaving a slight noise at the beginning.

**Steps to Reproduce**
1. Build and install `piu/sounds` example app using `mcconfig -m -p esp32/m5stack`.
2. The device cannot play any sound including startup sound(bflatmajor.maud) .

**Other information**
- The sound works fine on M5Stack Basic and Core2 with Moddable v4.2.1
- The issue seems similar to what is described in https://github.com/Moddable-OpenSource/moddable/issues/881. I am currently investigating whether this is a degradation on the esp-idf side or if there is something problematic in the CoreS3 initialization code that I wrote.

meganetaaan:
I've been progressing with debugging, but I still haven't seen a solution. As I continue to read through the esp-idf code, I'm starting to think that the cause might be the incorrect calculation of clock division, which was mentioned in a previous discussion.
https://github.com/Moddable-OpenSource/moddable/discussions/1134#discussioncomment-6165089

Indeed, regarding this calculation process, there have been changes from esp-idf v4 to v5, and it seems that even if the same sample rate is given, the results are different.
v5.1: https://github.com/espressif/esp-idf/blob/release/v5.1/components/hal/i2s_hal.c#L36
v4.4: https://github.com/espressif/esp-idf/blob/release/v4.4/components/hal/i2s_hal.c#L28

However, I'm not sure how to identify the cause from here or how to fix it, and I'm stuck. I would appreciate assistance from anyone with knowledge of esp-idf and who has the CoreS3.

phoddie:
Thank you for working on this issue. I hoped that IDF v5 would bring an end to these kinds of audio problems.

There isn't much to configure for the I2S audio output:
- The I2S pins must be correct (the hardware didn't change and it worked before!)
- It must still be 16-bit audio (right?)
- The M5Stack Core S3 is the **only** device that's manifest uses the `mck` pin (set to 0, which is a little suspicious). Is that really required? Could it actually be causing the problem somehow?
[...]

PR #1297, opened 2024-01-20
Switch slot of I2S

stc1988:
This PR is fix for #1236.

Setting I2S slot to left make M5Stack cores3 to AudioOut sound. 
This fix affects only `esp32/m5stack_cores3`.

### DIFF
--- a/build/devices/esp32/targets/m5stack_cores3/manifest.json
+++ b/build/devices/esp32/targets/m5stack_cores3/manifest.json
@@ -106,7 +106,7 @@
 			"volume_divider": 1,
 			"i2s": {
 				"num": 1,
-				"slot": 1,
+				"slot": "I2S_STD_SLOT_LEFT",
 				"bitsPerSample": 16,
 				"bck_pin": 34,
 				"lr_pin": 33,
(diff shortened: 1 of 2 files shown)


phoddie:
This is reasonable. Before merging, I'd like to quickly consider a slightly simpler approach. The manifest could use the ESP-IDF constant instead of a number.

```
    "slot": "#I2S_STD_SLOT_LEFT"
```

Then you can write:

```c
	#ifndef MODDEF_AUDIOOUT_I2S_SLOT
 		#define MODDEF_AUDIOOUT_I2S_SLOT I2S_STD_SLOT_RIGHT
 	#endif
```

And 

```c
i2s_config.slot_cfg.slot_mask = MODDEF_AUDIOOUT_I2S_SLOT;
```

I think it is more clear. What do you think? (I'm OK with either approach.)

stc1988:
Thanks for the review.
I think that is more clear too.

I have committed to this fix.

phoddie:
Thanks for considering the proposal. Also, thank you for correctly my mistake with the "#". ;)

We'll get this merged.

PR #1300, opened 2024-01-29
Add fuzzilli builds on Linux

raphdev:
XS allows builds of xst that implement the Fuzzilli REPRL protocol for fuzzing by enabling the `FUZZING` and `FUZZILLI` flags. However, they were only available for Mac. This PR enables the corresponding Linux build, removing the requirement of a Mac machine for deeper engine fuzzing.

Aside from the fuzzilli coverage guards being disabled in Linux, the stack overflow check interferes with ASAN's `stack-use-after-free` due to promoting stack variables to the heap, which breaks the check's assumptions. This prevented fuzzilli on Linux from successfully starting and gaining coverage. To avoid this, we disable ASAN in these functions.

An alternative to disabling ASAN for this function could be using ASAN's interface to get the addr of the pointer on the stack, but I have not tested that yet.

### DIFF
--- a/xs/sources/xsMemory.c
+++ b/xs/sources/xsMemory.c
@@ -264,6 +264,12 @@ void* fxCheckChunk(txMachine* the, txChunk* chunk, txSize size, txSize offset)
 	return C_NULL;
 }
 
+#if defined(__clang__) || defined (__GNUC__)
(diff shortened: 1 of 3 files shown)


phoddie:
Merged

PR #1301, opened 2024-01-29
Add support for more sanitizers in fuzzilli builds

raphdev:
This PR builds on #1300 and adds additional sanitizers in debug builds:

* [UndefinedBehaviorSanitizer (UBSAN)](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html), which allows fuzzilli to find test cases that invoke undefined behavior. This can find issues that lead to the kind of memory violations that ASAN detects much earlier (e.g. optimized out bounds checking before it leads to an overflow).
* [MemorySanitizer (MSAN)](https://clang.llvm.org/docs/MemorySanitizer.html), which allows fuzzilli to find test cases that read uninitialized memory. Note that this sanitizer is only available in Linux.

These can be set via `SANITIZER=undefined` and `SANITIZER=memory` when building. The default for debug builds remains ASAN.

I have tested:

* These changes don't affect Mac builds (except for spurious stack overflow findings, which have occasionally shown up in Mac runs).
* UBSAN builds run on Mac by running for a couple weeks on an M1 Mac while I worked on the Linux port.
* UBSAN and MSAN builds run on Linux. They have run for weeks now and seem stable.

### DIFF
--- a/xs/makefiles/lin/xst.mk
+++ b/xs/makefiles/lin/xst.mk
@@ -41,8 +41,7 @@ SRC_DIR = $(XS_DIR)/sources
 TLS_DIR = $(XS_DIR)/tools
 TMP_DIR = $(BUILD_DIR)/tmp/lin/$(GOAL)/$(NAME)
 
-MACOS_ARCH ?= -arch i386
-MACOS_VERSION_MIN ?= -mmacosx-version-min=10.7
+ARCH := $(shell uname -m)
 LINK_OPTIONS = -rdynamic
 
 C_OPTIONS = \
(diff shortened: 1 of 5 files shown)


raphdev:
Some notes and food for thought:

* The UBSAN checks mirror oss-fuzz, minus floating-point division-by-zero checks (those will be disabled in the oss-fuzz builds soon). 


* I am open to another way to signal sanitizer choice. OSS-Fuzz does also set `SANITIZER`, but instead injects all sanitizer args via `CFLAGS`, but this requires another script/make target to set and track those accordingly. Embedding directly in xst's Makefile was the fastest way.

phoddie:
Merged

PR #1308, opened 2024-02-15
[typing]Outline.scale can omit parameter `y`

stc1988:


phoddie:
Strictly speaking it is a little different than omitting `y`. When there is a single argument, it Is used for both `x` and `y` scaling. This might be better expressed using function overloading.

```ts
    scale(scale: number): Outline
    scale(xScale: number, yScale: number): Outline
```

stc1988:
> Strictly speaking it is a little different than omitting `y`. When there is a single argument, it Is used for both `x` and `y` scaling. This might be better expressed using function overloading.
> 
> ```ts
>     scale(scale: number): Outline
>     scale(xScale: number, yScale: number): Outline
> ```

This is straightforward obviously. I committed your proposal. 
Thank you.

PR #1313, opened 2024-02-23
Mcu code

NW-Lab:
Hello

This is a PR for running mcu_code on Node-red-mcu.
https://github.com/phoddie/node-red-mcu/pull/122#issue-2150356003

Thank you.

### DIFF
--- a/tools/mcmanifest.js
+++ b/tools/mcmanifest.js
@@ -1946,8 +1946,14 @@ export class Tool extends TOOL {
 
 				flows.forEach((node, i) => {
 					let manifest;
-					if (node.moddable_manifest)
-						manifest = {...node.moddable_manifest};
+					if (node.moddable_manifest){
+						try{
+							manifest = JSON.parse(node.moddable_manifest);
+						}
+						catch{
+							manifest = {...node.moddable_manifest};
+						}
+					}
 					else if (nodeTypes[node.type])
 						manifest = {include: nodeTypes[node.type]};
 					else
--- a/tools/mcmanifest.js
+++ b/tools/mcmanifest.js
@@ -1946,13 +1946,11 @@ export class Tool extends TOOL {
 
 				flows.forEach((node, i) => {
 					let manifest;
-					if (node.moddable_manifest){
-						try{
+					if (node.moddable_manifest) {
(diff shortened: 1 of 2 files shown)


tools/mcmanifest.js
```diff
@@ -1946,8 +1946,14 @@ export class Tool extends TOOL {
 
 				flows.forEach((node, i) => {
 					let manifest;
-					if (node.moddable_manifest)
-						manifest = {...node.moddable_manifest};
+					if (node.moddable_manifest){
```
phoddie: My understanding is that this change was made to allow `moddable_manifest` to be a String containing JSON. As implemented, if there is a syntax error in the JSON, it will be ignored instead of reported. That will be confusing for developers. I suggest reworking the change as:

```js
	if (node.moddable_manifest) {
		if ("string" === typeof node.moddable_manifest)
			manifest = JSON.parse(node.moddable_manifest);
		else
			manifest = {...node.moddable_manifest};
	}
```

tools/nodered2mcu.js
```diff
@@ -470,7 +470,9 @@ export default class extends TOOL {
 				config.active = !!config.active;
 			} break;
 
-			case "function": {
+			case "function":
+			case "mcu_code":
+			{
```
phoddie: Since the "mcu_code" node is a duplicate of the "function", you can change the "mcu_code" to a "function" node here. That eliminates the need for "mcu_code" in Node-RED MCU.

```js
			case "mcu_code":
				config.type = "function";
				// fall through
			case "function": {
				const params = "node, context, flow, global, libs, env"
```

tools/mcmanifest.js
NW-Lab: It would be better if you make a suggestion.
thank you.

tools/nodered2mcu.js
NW-Lab: At this point, it would be best if you could make a suggestion.
When the function and mcu_code nodes have different functions, it is better to separate them.
thank you.

NW-Lab:
This PR has been withdrawn and replaced with the following PR.
https://github.com/Moddable-OpenSource/moddable/pull/1317

PR #1316, opened 2024-02-25
wavsteam for OpenAI text-to-speech

stc1988:
This PR addd parameter `response_format` to [OpenAI text-to-speech.](https://platform.openai.com/docs/api-reference/audio/createSpeech)
Depending on the situation, I think there are cases where `wav` might be better than `mp3`.

Furthermore, this PR includes change for the following points.

* split `manifest.json` of  `http-steam` into `manifest_sbcstreamer.json` and `manifest_wavstremer.json`
* TLS Certification for OpenAI API.

### DIFF
--- a/examples/pins/audioout/openai-stream/openaistreamer.js
+++ b/examples/pins/audioout/openai-stream/openaistreamer.js
@@ -26,27 +26,21 @@ import WavStreamer from "wavstreamer";
 export default class {
 	constructor(options) {
 		const {key, model, voice, input, speed, response_format, ...o} = options;
-		const reuest = {
+		const request = {
 			input,
 			model,
 			voice,
 			speed: speed ?? 1,
 			response_format: response_format ?? "mp3",
 		};
-		const body = ArrayBuffer.fromString(JSON.stringify(reuest));
+		const body = ArrayBuffer.fromString(JSON.stringify(request));
 		let streamer;
-		if (reuest.response_format == "mp3") {
+		if (request.response_format == "mp3") {
 			streamer = MP3Streamer;
-		} else if (reuest.response_format == "wav") {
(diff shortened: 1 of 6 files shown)


examples/pins/audioout/openai-stream/openaistreamer.js
```diff
@@ -19,20 +19,37 @@
...
+import WavStreamer from "wavstreamer";
 
 // https://platform.openai.com/docs/api-reference/audio/createSpeech
 
 export default class {
 	constructor(options) {
-		const {key, model, voice, input, speed,  ...o} = options;
-		const body = ArrayBuffer.fromString(JSON.stringify({
+		const {key, model, voice, input, speed, response_format, ...o} = options;
+		const reuest = {
```
phoddie: `reuest` should be `request`

examples/pins/audioout/openai-stream/openaistreamer.js
```diff
@@ -19,20 +19,37 @@
...
+		} else if (reuest.response_format == "wav") {
+			streamer = WavStreamer;
+		} else {
+			const error = new Error(`invalid response_format:${reuest.response_format}`);
+			if (options.onError) {
+				options.onError(error);
+				return;
+			} else {
+				throw error;
+			}
```
phoddie: This should just be "throw new Error(`invalid response_format:${request.response_format}`)". That is because `onError` should not be called from the constructor. See ECMA-419 [Callbacks](https://419.ecma-international.org/#-8-base-class-pattern-callbacks): "callbacks may not be invoked by the instance from... the constructor."

phoddie (commented):
This looks good. There are just a couple little details to fix.

stc1988:
Thank you for review.
I fixed review comments

PR #1317, opened 2024-02-27
Node-RED-mcu  Load manifest and import from comments

NW-Lab:
Hello

This is a modification of https://github.com/Moddable-OpenSource/moddable/pull/1313.
Read manifest and import from comment nodes in Node-RED MCU.

For manifest, name it  “moddable_manifest” and write the content in the description.
```
{
    "include":"$(MODDABLE)/modules/drivers/aw3641/manifest.json"
}
```

For import, name it “moddable_import” and write the content in the description.
```
{
   "Variable name 1":"Module name 1",
   "Variable name 2":"Module name 2"
}
```
If the Function node code can only be executed by MCU, use RED.mcu
```
if(RED.mcu){
     code......
}
```

Sample flows.json
This is an example of running [moddable/examples/drivers/aw3641](https://github.com/Moddable-OpenSource/moddable/tree/public/examples/drivers/aw3641)  on Node-red MCU Edition.
image

```
[
    {
        "id": "a9469753d31dfb25",
        "type": "tab",
        "label": "sample flow",
        "disabled": false,
        "info": "",
        "env": [],
        "_mcu": {
            "mcu": true
        }
    },
    {
        "id": "4043457111bf2b23",
        "type": "inject",
        "z": "a9469753d31dfb25",
        "name": "inject",
        "props": [
            {
                "p": "payload"
            },
            {
                "p": "topic",
                "vt": "str"
            }
        ],
        "repeat": "",
        "crontab": "",
        "once": false,
        "onceDelay": 0.1,
        "topic": "",
        "payload": "",
        "payloadType": "date",
        "_mcu": {
            "mcu": true
        },
        "x": 130,
        "y": 140,
        "wires": [
            [
                "73b0486d25bfe6dc"
            ]
        ]
    },
    {
        "id": "c7d233f7729a0722",
        "type": "debug",
        "z": "a9469753d31dfb25",
        "name": "debug 1",
        "active": true,
        "tosidebar": true,
        "console": false,
        "tostatus": false,
        "complete": "false",
        "statusVal": "",
        "statusType": "auto",
        "_mcu": {
            "mcu": true
        },
        "x": 440,
        "y": 140,
        "wires": []
    },
    {
        "id": "1298a2c40c7497d1",
        "type": "comment",
        "z": "a9469753d31dfb25",
        "name": "moddable_manifest",
        "info": "{\n    \"include\":\"$(MODDABLE)/modules/drivers/aw3641/manifest.json\"\n}",
        "_mcu": {
            "mcu": true
        },
        "x": 150,
        "y": 60,
        "wires": []
    },
    {
        "id": "73b0486d25bfe6dc",
        "type": "function",
        "z": "a9469753d31dfb25",
        "name": "function 2",
        "func": "msg.payload = \"abc\";\nreturn msg;",
        "outputs": 1,
        "timeout": 0,
        "noerr": 0,
        "initialize": "if(RED.mcu){\n    //const val=AW3641.off;\n    //const val=AW3641.Time220ms_Brightness100;\n    //const val=AW3641.Time220ms_Brightness90;\n    //const val=AW3641.Time220ms_Brightness80;\n    //const val=AW3641.Time220ms_Brightness70;\n    //const val=AW3641.Time220ms_Brightness60;\n    const val = AW3641.Time220ms_Brightness50;\n    //const val=AW3641.Time220ms_Brightness40;\n    //const val=AW3641.Time220ms_Brightness30;\n    //const val=AW3641.Time1_3s_Brightness100;\n    //const val=AW3641.Time1_3s_Brightness90;\n    //const val=AW3641.Time1_3s_Brightness80;\n    //const val=AW3641.Time1_3s_Brightness70;\n    //const val=AW3641.Time1_3s_Brightness60;\n    //const val=AW3641.Time1_3s_Brightness50;\n    //const val=AW3641.Time1_3s_Brightness40;\n    //const val=AW3641.Time1_3s_Brightness30;\t\n\n    const lamp = new AW3641({ pin: 26, });\n\n    Timer.repeat(() => {\n        lamp.flash(val);\n        trace(\"flash\\n\");\n    }, 2000);\n}",
        "finalize": "",
        "libs": [
            {
                "var": "Timer",
                "module": "timer"
            }
        ],
        "_mcu": {
            "mcu": true
        },
        "x": 280,
        "y": 140,
        "wires": [
            [
                "c7d233f7729a0722"
            ]
        ]
    },
    {
        "id": "27ca19e76c477ddb",
        "type": "comment",
        "z": "a9469753d31dfb25",
        "name": "moddable_import",
        "info": "{\n    \"Timer\":\"timer\",\n    \"AW3641\":\"aw3641\"\n}",
        "_mcu": {
            "mcu": true
        },
        "x": 400,
        "y": 60,
        "wires": []
    }
]
```

Thank you.

### DIFF
--- a/tools/mcmanifest.js
+++ b/tools/mcmanifest.js
@@ -1946,9 +1946,7 @@ export class Tool extends TOOL {
 
 				flows.forEach((node, i) => {
 					let manifest;
-					if(("comment" === node.type)&&("moddable_manifest" === node.name)&&(node.info?.startsWith("{"))){
-						trace("*** Manifest addition by comment ***\n");
-						trace(node.info+"\n");
+					if(("comment" === node.type)&&(node.name?.startsWith("moddable_manifest"))&&(node.info?.startsWith("{"))){
 						manifest = JSON.parse(node.info);
 					}
 					else if (node.moddable_manifest)
--- a/tools/mcmanifest.js
+++ b/tools/mcmanifest.js
@@ -1946,7 +1946,12 @@ export class Tool extends TOOL {
 
 				flows.forEach((node, i) => {
 					let manifest;
-					if (node.moddable_manifest)
+					if(("comment" === node.type)&&("moddable_manifest" === node.name)&&(node.info?.startsWith("{"))){
+						trace("*** Manifest addition by comment ***\n");
+						trace(node.info+"\n");
+						manifest = JSON.parse(node.info);
+					}
+					else if (node.moddable_manifest)
 						manifest = {...node.moddable_manifest};
 					else if (nodeTypes[node.type])
 						manifest = {include: nodeTypes[node.type]};
--- a/tools/nodered2mcu.js
+++ b/tools/nodered2mcu.js
@@ -190,7 +190,6 @@ export default class extends TOOL {
 			if ("tab" === config.type)
 				continue;
 			if(("comment" === config.type)&&(config.name?.startsWith("moddable_import"))&&(config.info?.startsWith("{"))){
-				trace("*** Imports addition by comment ***\n");
 				const arr = JSON.parse(config.info);
 				for(let key in arr) {
 					imports.set(key, arr[key]);
--- a/tools/nodered2mcu.js
+++ b/tools/nodered2mcu.js
@@ -189,6 +189,14 @@ export default class extends TOOL {
 			const config = flows[i];
 			if ("tab" === config.type)
 				continue;
+			if(("comment" === config.type)&&("moddable_import" === config.name)&&(config.info?.startsWith("{"))){
+				trace("*** Imports addition by comment ***\n");
+				const arr = JSON.parse(config.info);
+				for(let key in arr) {
+					trace("import "+key +" from " + arr[key] +"\n");
+					imports.set(key, arr[key]);
+				}
+			}
 			if (config.d || ("comment" === config.type)) {
 				flows.splice(i, 1);
 				i -= 1;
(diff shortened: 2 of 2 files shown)


phoddie:
This is an elegantly simple solution to the problem we've been discussing. While it isn't the most developer-friendly solution, it is a great way to experiment further. We should consider this an experimental feature that might eventually be replaced.

There are two changes I'd like to request before merging your PR:

1. `nodered2mcu` does not usually trace its progress to the console. Would you please remove the traces? Otherwise the output can become noisy.
2. It seems possible that a project might want more than one manifest, or more than one import. To keep those organized, they should have different names. Instead of checking the node name is "moddable_manifest" or "moddable_import", the check could use `startsWith` for those two strings. That way you can have "moddable_manifest aw3641", for example. This would be nice for reusing nodes in different flows.

NW-Lab:
@phoddie san

I tried to fix it. Is this okay?

I attached trace because it can also be debugged. I deleted it because it is unnecessary.
It's a very good idea to only look at the first part of the comment's name. This is important for the Node-RED ecosystem.

Thank you.

phoddie:
> I tried to fix it. Is this okay?

Looks great!

> I attached trace because it can also be debugged. I deleted it because it is unnecessary.

I sometimes commit traces from my debugging / development too. Not a big deal.

> It's a very good idea to only look at the first part of the comment's name. This is important for the Node-RED ecosystem.

Thank you for being open to the suggestion.

We'll get this merged soon.

phoddie:
This should be merged as part of Moddable SDK 4.5. Please confirm. Thank you.

NW-Lab:
I think that it is all right.
thank you.

PR #1320, opened 2024-03-08
fix: don't crash when snapshotting a host object that has a (valid) destructor

FUDCo:
If you create a host object using `xsNewHostObject` and provide it with a (valid) destructor, attempting to perform a snapshot causes a crash.   I think this was a simple oversight, but in any case this is a proposed fix.

### DIFF
--- a/xs/sources/xsSnapshot.c
+++ b/xs/sources/xsSnapshot.c
@@ -1092,10 +1092,7 @@ void fxMeasureSlot(txMachine* the, txSnapshot* snapshot, txSlot* slot, txSize* c
 				destructor = slot->value.host.variant.hooks->destructor;
 			else
 				destructor = slot->value.host.variant.destructor;
-		    if (dladdr(destructor, &info)) {
-				mxAssert(0, "# snapshot: no host destructor: %s!\n", info.dli_sname);
-			}
-			else {
+		    if (dladdr(destructor, &info) == 0) {
 				mxAssert(0, "# snapshot: no host destructor!\n");
 			}
 		}


phoddie:
Interesting. Thanks for the report! As a rule, host objects cannot go into a snapshot because there's no way to guarantee they are reloaded in a compatible environment. Of course, it shouldn't crash. 

Would you share a test case that we can use to better understand?

FUDCo:
The test case is rather contrived as I was creating a simple proof-of-concept exercise just to make sure I could correctly follow the whole process end-to-end of creating a JS class in C code, then getting it to build, link, and run correctly (this as preamble to a much more ambitious project that I didn't want to get entangled with these issues, namely porting (or, perhaps, more accurately given the vast API differences between Node and XS, rewriting) the npm better-sqlite package to allow SQLite to be used from code running in XS -- BTW, the XS native code API is **so** much more straightforward than Node's).

The constraints on snapshotting a heap containing host objects seem to be a bit weird.  If the host objects in question have no host data it doesn't seem to be a problem -- xsnap has several of these essentially stateless host objects (notably `TextEncoder` and `TextDecoder`) and everything there seems to work fine.  However, my read of the XS code is that if the host data in question doesn't contain any pointers things or things that indirectly refer to pointers it should also work OK.  Of course, it doesn't appear to me like there's a graceful way to automatically ensure one's host data is actually "pure" in this way; you just have to understand your code really well.  The particular things that look tricky to me are file descriptors and `FILE *` pointers, which are relevant to the use case I'm anticipating.  It might be better if host objects left their host data behind on snapshot and always got restored from snapshot with their host data as NULL.  In this case you could regard the host data as a kind of cache that would need to be reconstructed when you notice there's a NULL there; this would work for my use case (where the pathname associated with the file descriptor, stored in the JS state, could be used to reopen the file) though I'm not deeply confident of how well this would generalize.  As is I think I'll have to keep track of when I'm reloading from a snapshot and force the issue, which probably means actually having a stateless host object and a lookup table outside the host objects that is guaranteed not to survive restart.

For the record, the test case I wrote is:
```c
#include <stdlib.h>
#include "xs.h"

void xs_nativeTest_destructor(void *data) {
    if (data) {
        free(data);
    }
}

void xs_doit(xsMachine *the) {
    xsSlot jsTag = xsString(xsGetHostData(xsThis));
    xsResult = xsCall2(jsTag, xsID("concat"), xsString(" "), xsArg(0));
}

void xs_NativeTest(xsMachine *the) {
    xsVars(1);
    xsVar(0) = xsGet(xsTarget, xsID("prototype"));
    xsResult = xsNewHostInstance(xsVar(0));
    xsThis = xsResult;
    char *rawTag = xsToString(xsArg(0));
    char *tag = (char *) malloc(strlen(rawTag) + 1);
    strcpy(tag, rawTag);
    xsSetHostData(xsThis, tag);
}

void installNativeTest(xsMachine *the) {
#define kPrototype (0)
#define kConstructor (1)
#define kScratch (2)

    xsBeginHost(the);
    xsVars(3);

    xsVar(kPrototype) = xsNewHostObject(xs_nativeTest_destructor);
    xsVar(kConstructor) = xsNewHostConstructor(xs_NativeTest, 1, xsVar(kPrototype));
    xsDefine(xsGlobal, xsID("NativeTest"), xsVar(kConstructor), xsDontEnum);

    xsVar(kScratch) = xsNewHostFunction(xs_doit, 1);
    xsDefine(xsVar(kPrototype), xsID("doit"), xsVar(kScratch), xsDefault);

    xsEndHost(the);
}
```
And, of course, in xsnap the `gxSnapshotCallbacks` table needed to have `xs_NativeTest` and `xs_doit` added at the end and to call `installNativeTest ` from `xsBuildAgent`.  It was `xs_nativeTest_destructor` that tripped up the code in `fxMeasureSlot` that's the subject of this PR.  With the proposed change in place, the snapshot operation seems to run fine, though I haven't attempted to build a test to see if restoring this works as I'm increasingly thinking I'm going to have to go with the stateless host object approach for my ultimate goal anyway.

phoddie:
Thanks for the details and notes.

The fundamental reality is that host data cannot be safely stored to a snapshot.  Pure host data is just a pointer -- there's not even a byte length -- so it cannot be stored at all. As an exception, host data stored in a chunk may be stored in a snapshot. XS knows the byte length of those. This is what `TextDecoder` uses. As you note, it is the responsibility of the author of the host object to ensure that content of the host chunk is endian-independent data (no pointers, runtime objects, etc).

> It might be better if host objects left their host data behind on snapshot and always got restored from snapshot with their host data as NULL

That would require every host method to work when host data is NULL which feels fragile. Also, for `TextDecoder` it wouldn't work as the stored state is necessary for correct execution.

> For the record, the test case I wrote is...

So, creating an instance of `NativeTest` and then creating a snapshot will crash? We should fix that. (It will still fail, but it should fail more gracefully

> BTW, the XS native code API is so much more straightforward than Node's

Thanks! We hear that pretty often from the small group of people who have experience with native APIs across engines. ;)

FUDCo:
> The fundamental reality is that host data cannot be safely stored to a snapshot. Pure host data is just a pointer -- there's not even a byte length -- so it cannot be stored at all. As an exception, host data stored in a chunk may be stored in a snapshot. XS knows the byte length of those. This is what `TextDecoder` uses. As you note, it is the responsibility of the author of the host object to ensure that content of the host chunk is endian-independent data (no pointers, runtime objects, etc).

Ah, I think I missed the subtleties of the host data vs. host chunk distinction.  I was principally cribbing from `TextEncoder` which has neither host data nor host chunk and seems to be purely computational.  I could probably rewrite my test example to use a host chunk instead and bypass this issue entirely.  Though at this point I've got the build-link-execute pathway paved so there'd be no point.

> > It might be better if host objects left their host data behind on snapshot and always got restored from snapshot with their host data as NULL
> 
> That would require every host method to work when host data is NULL which feels fragile. Also, for `TextDecoder` it wouldn't work as the stored state is necessary for correct execution.

In my defense, I **did** say I wasn't confident of this approach :-)

> So, creating an instance of `NativeTest` and then creating a snapshot will crash? We should fix that. (It will still fail, but it should fail more gracefully

Yes.  The code itself fails via an assert at the point which this PR edits, but the assert message (which would have been plenty confusing enough if it had actually been output) seemed to get swallowed somewhere up the stack so what I saw from xsnap was `cannot write snapshot @7: Invalid argument`, which was unsurprisingly baffling.

Based on your explanation (which totally makes sense to me), I suspect it should simply reject any attempt to snapshot an object with host data at all with a "you can't do that" message, and never mind the business with the destructor.

phoddie:
Yeah, OK. So, I think the constraints are still appropriate (e.g. host data pointers cannot be written to a snapshot). The crash needs to be fixed. We can take a look at doing that. The error message reported isn't optimal -- though the error generated in xsSnapshot.c, "snapshot: no host destructor" or "snapshot: no host data," seem to point you in the right direction.

phoddie:
The bug causing the crash you see is here:

https://github.com/Moddable-OpenSource/moddable/blob/c978d429a9dfc148a5cedf0ff470ad98e9adfc94/xs/sources/xsSnapshot.c#L1164

xsSnapshot wants to display the name of the destructor in the error message, if possible. Even when `dladdr` succeeds it doesn't always provide the name. So the fix is to check that the name is non-NULL before using it:

```c
	 if (dladdr(destructor, &info) && info.dli_sname) {
```

We'll apply that to this repository. (It occurs in two places in xsSnapshot.c and will be corrected in both.)

Thanks again for reporting this. Glad we had a chance to talk through how native host implementations interact with snapshots.

PR #1323, opened 2024-03-13
No ASAN on regex stack check

raphdev:
XS checks by stack overflow by checking if new stack variables have a lower address than the one returned by `fxCStackLimit`, stored in the VM's `stackLimit` or in this case the parser structure's `stackLimit`. However, ASAN can change the memory layout in ways that break this check (see #1300 for more context). We disable ASAN in this function, to prevent instrumentation from interfering with stack overflow detection.


phoddie:

raphdev:
This was merged in https://github.com/Moddable-OpenSource/moddable/commit/5373fcce0ba239c5bbf4005e5ca76e837a25af0f

PR #1324, opened 2024-03-20
Add support for M5Stack Core2 v1.1

meganetaaan:
This is a pull request to support the minor update version v1.1 of the M5Stack Core2.
In v1.1, the power management IC has been changed from AXP192 to AXP2101 (which is also used in M5Stack CoreS3).

- In the setup-target.js of m5stack_core2, the power management IC's id is read and automatically identified. This maintains compatibility with previous versions of the M5Stack Core2.
 - Added some functions that are present in AXP192 but not in AXP2101.
- Some methods, such as setChargeEnable, reported by another user as not functioning, have been temporarily set to a not implemented status. We consider this to be of low urgency and would like to address it in a separate PR.

This change has been tested by writing the piu/balls demo to M5Stack v1.0, v1.1, and M5Stack CoreS3, and confirming its operation.

This PR was created with the cooperation of a new contributor, @RChikamura . He has been instructed to submit a CLA.

### DIFF
--- a/build/devices/esp32/targets/m5stack_core2/setup-target.js
+++ b/build/devices/esp32/targets/m5stack_core2/setup-target.js
@@ -19,12 +19,14 @@
  */
  
 import AXP192 from "axp192";
+import AXP2101 from "axp2101";
 import MPU6886 from "mpu6886";
 import AudioOut from "pins/audioout";
 import Resource from "Resource";
 import Timer from "timer";
 import config from "mc/config";
 import I2C from "pins/i2c";
+import SMBus from "pins/smbus";
 
 const INTERNAL_I2C = Object.freeze({
 	sda: 21,
@@ -70,7 +72,14 @@ export default function (done) {
 	};
 
 	// power
-	globalThis.power = new Power();
+  debugger;
+  const powerICID = new SMBus({
+    ...INTERNAL_I2C,
+    address: 0x34,
+  }).readByte(0x03)
+
+	globalThis.power =
+    powerICID === 0x03 ? new PowerAXP192() : new PowerAXP2101() /* expects 0x4a */;
 
 	// speaker
 	power.speaker.enable = true;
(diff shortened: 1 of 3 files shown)


build/devices/esp32/targets/m5stack_core2/setup-target.js
```diff
@@ -70,7 +72,14 @@ export default function (done) {
 	};
 
 	// power
-	globalThis.power = new Power();
+  debugger;
```
phoddie: Please remove the `debugger` statement.

phoddie:
Welcome to @RChikamura! Thank you both for the PR. After we receive the CLA, we will look at merging this.

build/devices/esp32/targets/m5stack_core2/setup-target.js
meganetaaan: (Oops) fixed!

build/devices/esp32/targets/m5stack_core2/setup-target.js
```diff
@@ -70,7 +72,13 @@ export default function (done) {
 	};
 
 	// power
-	globalThis.power = new Power();
+  const powerICID = new SMBus({
```
phoddie: While this code does work, it would be best to immediately close the SMBus instance rather than waiting for it to be garbage collected. I suggest the following:

```js
  const s = new SMBus({
     ...INTERNAL_I2C,
     address: 0x34,
   })
  const powerICID = s.readByte(0x03)
  s.close()
```

phoddie: @meganetaaan - What do you think about this change? If you can make the adjustment, then we should be able to merge this. Thank you!

meganetaaan: Much better! I confirmed that the fix works then pushed.

phoddie: Great, thanks for making the change.

phoddie:
This has been [merged](https://github.com/Moddable-OpenSource/moddable/commit/fc5c2b41c0e2df4f9235f11aa1962886a603d4dd). Thank you!

PR #1340, opened 2024-04-30
Apply memory limits during fuzzing

raphdev:
This PR enables the newly added `mxXSMemoryLimit` that tracks `mxNoChunks` allocator and aborts the machine when the provided limit is exceeded. It allows us to gracefully handle fuzzer test cases to avoid reporting JS code that allocates beyond the fuzzer limits as bugs.

In fuzzing builds, we also use a function that purges all memory that we call on abort paths. This makes previous work to tune ASAN's leak sanitizer unnecessary (#1067). We removed it as it had no effect.


raphdev:
Closing as it was committed in 7efb104c88ea4a02fb3ad9b035f9f644316861f5 and e97b4526671ee09a808af83b53366d131b6da24c.

PR #1343, opened 2024-05-06
Add WiFi event definition  "station_connect" and "station_disconnect"

stc1988:
In discussion https://github.com/Moddable-OpenSource/moddable/discussions/1342, I use motitoring msg `station_connect` and `station_disconnect` to detect connected devices in Wifi AP mode.

This PR add these definitions to `Wifi` class.


phoddie:
Thanks for getting this started. Since these callbacks are only available on ESP32, the constants should be added to the [ESP32 version of wifi.js](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/wifi/esp32/wifi.js). (The wifi.d.ts change should still be OK)

stc1988:
Thanks for reviewing. I moved the constants to ESP32 version.

PR #1345, opened 2024-05-13
add typing qrcode

stc1988:
This PR add typings for `modules/data/qrcode` and `drawQRCode`.
And I found that there was an unnecessary parameter `back` in `drawQRCode`, so I corrected it.

### DIFF
--- a/modules/commodetto/qrcode/pocoqrcode.js
+++ b/modules/commodetto/qrcode/pocoqrcode.js
@@ -20,5 +20,5 @@
 
 import Poco from "commodetto/PocoCore";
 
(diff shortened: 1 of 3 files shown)


phoddie:
> And I found that there was an unnecessary parameter back in drawQRCode, so I corrected it

Good catch, thank you. It looks like that is a remnant from an earlier version of the code.

phoddie:
[Merged](https://github.com/Moddable-OpenSource/moddable/commit/424d1797975a2a603b4a17385fd675589a4e7400). Thank you!

PR #1347, opened 2024-05-15
Conditional signatures fxNewHostConstructor, fxNewHostFunction and fxNewHostInstance with -DINCLUDE_XSPLATFORM

SMotaal:
This PR adds conditional signatures with `-DINCLUDE_XSPLATFORM` to return `xsSlot*` instead of `void` for `fxNewHostConstructor`, `fxNewHostFunction` and `fxNewHostInstance`.

The current signatures result in conflicts when building xsnap with Emscripten, which are mitigated by the proposed changes. See [@SMotaal/xs-wasm](https://github.com/SMotaal/xs-wasm/tree/main?tab=readme-ov-file#patching) for more.

I am not sure if the proposed changes align with other build paths, and would appreciate your input to avoid breaking existing projects.

### DIFF
--- a/xs/includes/xs.h
+++ b/xs/includes/xs.h
@@ -1439,9 +1439,15 @@ mxImport void fxArrayCacheEnd(xsMachine*, xsSlot*);
 mxImport void fxArrayCacheItem(xsMachine*, xsSlot*, xsSlot*);
 
 mxImport void fxBuildHosts(xsMachine*, xsIntegerValue, xsHostBuilder*);
+#ifndef INCLUDE_XSPLATFORM
 mxImport void fxNewHostConstructor(xsMachine*, xsCallback, xsIntegerValue, xsIntegerValue);
 mxImport void fxNewHostFunction(xsMachine*, xsCallback, xsIntegerValue, xsIntegerValue, xsIntegerValue);
 mxImport void fxNewHostInstance(xsMachine*);
+#else
+mxImport xsSlot* fxNewHostConstructor(xsMachine*, xsCallback, xsIntegerValue, xsIntegerValue);
+mxImport xsSlot* fxNewHostFunction(xsMachine*, xsCallback, xsIntegerValue, xsIntegerValue, xsIntegerValue);
+mxImport xsSlot* fxNewHostInstance(xsMachine*);
+#endif
 mxImport xsSlot* fxNewHostObject(xsMachine*, xsDestructor);
 mxImport xsIntegerValue fxGetHostBufferLength(xsMachine*, xsSlot*);
 mxImport void* fxGetHostChunk(xsMachine*, xsSlot*);


phoddie:
Interesting. I'm not sure why those functions are declared inconsistently. It seems like it would be better for them to always have the same signature, the one that matches the implementation. To understand if that's possible, they would need to be checked across the different build targets.

I'm curious: is there a particular reason you conditionalized the change on `INCLUDE_XSPLATFORM `?

SMotaal:
I've been running into this problem with `emcc` specifically for the 9 months I've been working on this:

```
wasm-ld: error: function signature mismatch: fxNewHostFunction
>>> defined as (i32, i32, i32, i32, i32) -> void in tmp/wasm/debug/xsnap-emscripten/textdecoder.o
>>> defined as (i32, i32, i32, i32, i32) -> i32 in tmp/wasm/debug/xsnap-emscripten/xsAPI.o

wasm-ld: error: function signature mismatch: fxNewHostInstance
>>> defined as (i32) -> void in tmp/wasm/debug/xsnap-emscripten/textdecoder.o
>>> defined as (i32) -> i32 in tmp/wasm/debug/xsnap-emscripten/xsAPI.o

wasm-ld: error: function signature mismatch: fxNewHostConstructor
>>> defined as (i32, i32, i32, i32) -> void in tmp/wasm/debug/xsnap-emscripten/textdecoder.o
>>> defined as (i32, i32, i32, i32) -> i32 in tmp/wasm/debug/xsnap-emscripten/xsAPI.o
emcc: error: 'xs-wasm/upstream/@emscripten-core/emsdk/upstream/bin/wasm-ld @/var/folders/4k/yply4l3j44gbj0cpj3w0s0800000gp/T/emscripten_50axiv2r.rsp.utf-8' failed (returned 1)
```

For context, I use [xsnap-emscripten.mk](https://github.com/SMotaal/xs-wasm/blob/main/makefiles/wasm/xsnap-emscripten.mk) which was derived from [xsnap.mk](https://github.com/agoric-labs/xsnap-pub/blob/Agoric/xsnap/makefiles/mac/xsnap.mk).

I'm not sure why the signatures are inconsistent, and the same sources build natively on mac without any problems. With limited insights into the problem, this change seemed reasonable enough and borrows on the same from [xs.h#L41](https://github.com/Moddable-OpenSource/moddable/blob/public/xs/includes/xs.h#L41).

Your insights would be appreciated.

phoddie:
That's interesting. C calling conventions on a register-based instruction set (basically everything) wouldn't care about the difference because the return value comes back in a register. But...  mismatch would a problem for WASM because it is purely stack based (I think....). That said, we have a WASM build of XS (and many of its tools and a chunk of the Moddable SDK) and haven't run into this problem.

We'll look into just making the declarations consistent everywhere. That seems to make the most sense, at least at the moment.

SMotaal:
I agree, making the declarations consistent would be greatly appreciated. It's the ideal path instead of the workaround in this PR — thank you 🙌

phoddie:
Oops.... we will make that change. Thanks again.

phoddie:
[Changed applied](https://github.com/Moddable-OpenSource/moddable/commit/798de30d21365c6f243dd727ec3e51ce4afa6185).

PR #1351, opened 2024-06-02
onerror callback of eventsource  sets http status code

stc1988:
I am implementing the [streaming response of the ChatGPT API](https://platform.openai.com/docs/api-reference/streaming) using EventSource, but with the current EventSource, I cannot check the reason for the error response from the API. 

Therefore, in this Pull Request, I will set the `status` and `statusText` in the `onerror` callback when there is an error response from the API with a status other than 200.

### DIFF
--- a/examples/io/tcp/eventsource/eventsource.js
+++ b/examples/io/tcp/eventsource/eventsource.js
@@ -24,6 +24,51 @@ import Timer from "timer";
 import Headers from "headers";
 import URL from "url";
 
+const statusTexts = {
+	100: "Continue",
+	101: "Switching Protocols",
+	200: "OK",
+	201: "Created",
+	202: "Accepted",
+	203: "Non-Authoritative Information",
+	204: "No Content",
+	205: "Reset Content",
+	206: "Partial Content",
+	300: "Multiple Choices",
+	301: "Moved Permanently",
+	302: "Found",
+	303: "See Other",
+	304: "Not Modified",
+	305: "Use Proxy",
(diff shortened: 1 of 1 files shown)


phoddie:
Thank you. This looks reasonable.

I have one question. I don't see `status` or `statusText` defined in the SSE specification. In fact, under [Implementation Advice](https://html.spec.whatwg.org/multipage/server-sent-events.html#implementation-advice) it says this:

> Implementations are especially encouraged to report detailed information to their development consoles whenever an [error](https://html.spec.whatwg.org/multipage/indices.html#event-error) event is fired, since little to no information can be made available in the events themselves

That almost suggests that the status information should **not** be provided. But that seems strange.

If possible, I would prefer not to invent something new but to use the properties that are common in existing implementations. Is there a precedent for `status` and `statusText` or something like that?

stc1988:
Thank you for the confirmation.

I implemented it with reference to [Eventsource client](https://github.com/EventSource/eventsource), but indeed it was an extended one.
The specification does not specifically mention Event, and it seems to set a generic Event object.
If anything, [CloseEvent](https://websockets.spec.whatwg.org/#the-closeevent-interface)might be close.

phoddie:
Thanks for the information. Being consistent EventSource client you reference is a good start. Reading that module's code reminded me that `status` and `statusText` are properties from `fetch()`. That all makes me comfortable with your design choices. We'll merge this.

Your code is also a reminder that we should **eventually** move the `statusTexts` to a common module so it doesn't need to be duplicated.

PR #1357, opened 2024-06-15
Merge all tsconfig properties, not just compilerOptions

cmidgley:
The manifest system allows for TypeScript settings `manifest.json` files, and merges the various`typescript.tsconfig.compilerOptions` settings into a final `tsconfig.json`.  This pull expands on that feature to merge all of the `typescript.tsconfig` section (not just `typescript.tsconfig.compilerOptions`).  

This is driven by a use case where a TypeScript transformer expects custom settings in `tsconfig.json` such as:

```json
{
    "compilerOptions": {
        "various": "typescript setting are in here"
    },
    "newTransformerTool": {
        "third-party": "options can go in here"
    }
}
```

This is a common technique used in the Node world for setting up build tools.

### DIFF
--- a/tools/mcmanifest.js
+++ b/tools/mcmanifest.js
@@ -2073,7 +2073,7 @@ export class Tool extends TOOL {
 	mergeProperties(targets, sources, exclude) {
 		if (sources) {
 			for (let name in sources) {
-				if (exclude && exclude.includes(name))
+				if (exclude?.includes(name))
 					continue;
 				let target = targets[name];
 				let source = sources[name];
--- a/tools/mcmanifest.js
+++ b/tools/mcmanifest.js
@@ -1051,6 +1051,7 @@ export class TSConfigFile extends FILE {
 	}
 	generate(tool) {
 		let json = {
+			...tool.typescript.tsconfig,
 			compilerOptions: {
 				baseUrl: "./",
 				forceConsistentCasingInFileNames: true,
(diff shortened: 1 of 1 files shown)


tools/mcmanifest.js
```diff
@@ -2068,9 +2070,11 @@ export class Tool extends TOOL {
 		}
 		return;
 	}
-	mergeProperties(targets, sources) {
+	mergeProperties(targets, sources, exclude) {
 		if (sources) {
 			for (let name in sources) {
+				if (exclude && exclude.includes(name))
```
phoddie: tiny nit: this could be simplified to `if (exclude?.includes(name))`

phoddie:
This looks OK. I understand that there are other properties that you may want to configure. The merging of `compilerOptions` is getting increasingly tricky. We'll get this merged (may be a few days though).

tools/mcmanifest.js
cmidgley: > tiny nit: this could be simplified to `if (exclude?.includes(name))`

no problem - that's the right way to do it.  change made.

cmidgley:
I agree the merge is getting complicated and open ended.  I just ran into another merge problem, so hold off on taking this PR while I work this.

phoddie:
OK, will wait. Merging is a tough problem. (The PR isn't horribly, terribly complicated.... but would be a little challenging to document what's going on.

cmidgley:
I found a work-around to the merge issue I was having, so this is good to go.

As for documenting what's going on, perhaps reference the `tsconfig.json.extends` property, as this change makes the merge quite similar to how it works (merging all of tsconfig, not just `compilerOptions`)?  Key differences are that Moddable has multiple parents (`extends` has just one), and that Moddable merges some arrays whereas tsc does not (for example, `types` arrays are merged in Moddable but not in tsc).

phoddie:
[Merged.](https://github.com/Moddable-OpenSource/moddable/commit/ee797bce86453c4cc93615e8a95c792519c94896)

PR #1364, opened 2024-07-04
Allow custom implementations of LogMachine with xsbug_log

cmidgley:
This PR is a proposal to add a tiny bit of extensibility to `xsbug_log` by allowing the `LogMachine` class to be overridden by an external module.  Changes include:

* Pulled `LogMachine` out to it's own file `xsbug-logmachine`
* Added check if the `XSBUG_LOGMACHINE` environment variable is set and if so load that as the module for `LogMachine` (else defaults to the original version)
* Bit of documentation in the associated xsbug readme

I am requesting this change for a couple of reasons:

1) Support for colorization using the \<info>, \<warn>, ... tags
    > I'm glad to provide this if interested, either as an update to the default `LogMachine` or as an optional one such as `xsbug-colorlogmachine.js` that can be selected with `XSBUG_LOGMACHINE`
2) I have a custom test runner, similar in concept to what `xsbug` does but unique to fit my needs, and I'm using `xsbug-log` to manage the logging and tracking of test executions.

### DIFF
--- a/documentation/tools/tools.md
+++ b/documentation/tools/tools.md
@@ -86,6 +86,8 @@ mcconfig [manifest] [-d] [-f format] [-i] [-m] [-o directory] [-p platform] [-r 
 
 > **Note**: The `-l` and `-dl` options requires Node.js on your build system. You must also first run `npm install` in `$MODDABLE/tools/xsbug-log`.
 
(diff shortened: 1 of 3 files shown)


phoddie:
It is great that you are using the xsbug API for your test runner. I always thought that API would be useful for automation.

It feels a little strange to me to build an extension mechanism into `xsbug-log`. The implementation is so simple that you could just replace `xsbug-log.js` and use the same supporting modules. But, I understand that the extension mechanism helps to avoid some code duplication and simplifies integration with `mcconfig` and `mcrun`.

We'll get this merged.

> Support for colorization using the <info>, <warn>, ... tags
> I'm glad to provide this if interested, either as an update to the default LogMachine or as an optional one such as xsbug-colorlogmachine.js that can be selected with XSBUG_LOGMACHINE

This would be welcome. In the interest of simplicity, It seems fine to just update the default rather than providing an alternate.

mkellner:

PR #1371, opened 2024-07-17
fix sample code of Neostrand documentation

stc1988:
This PR fixs some sample code of Neostrand documentation.

* `brightness` is not setter function, is property.
* remove invalid closing parentheses.


phoddie:
Thank you for the corrections! We will get this integrated.

mkellner:
Merged. It will be available in the next drop.

stc1988:
@phoddie  @mkellner 
Thank you for merge.

Additionally, can I confirm if `changeScheme` is incorrect?”
https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/drivers/neostrand/neostrand.md?plain=1#L167-L171

PR #1375, opened 2024-07-20
Fix typo in documentation/commodetto/poco.md

tenugui-taro:
Hello.

There appears to be an error in the documentation of the `fillPattern` function regarding the argument names. The previous notation repeated `sx` three times, but the correct arguments should be `sx, sy, sw, sh`.

Note: This message has been created using machine translation.

Thank you.

### DIFF
--- a/documentation/commodetto/poco.md
+++ b/documentation/commodetto/poco.md
@@ -658,7 +658,7 @@ The optional `blend` argument applies an additional blend level to all pixels in
 
 ***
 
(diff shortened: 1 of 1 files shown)


phoddie:
Copy and Paste are dangerous features. ;)

Thank you for your careful attention to detail while reviewing the translation. We'll get this merged.

mkellner:
This has been merged and will be available in the next update.

PR #1378, opened 2024-07-24
syntax highlighting for code block

stc1988:
In my translation work, I found that some code blocks do not have syntax highlighting set.　I believe that setting syntax highlighting will make the document easier to read and will also help to notice mistakes in the code blocks.

Therefore, there might be some parts where corrections were missed, but I made corrections as much as possible.


phoddie:
This is really great – thank you!

There are a couple of issues with the original JSON.

1. A few of them use `/* comments like this */` to provide instructions. Standard JSON has no comments so these are a syntax error. But, the comments are useful so maybe they should stay as-is?
2. The XS JSON parser used to permit a trailing comma (`{"foo": 1,}`) but no longer does because it is non-standard. Some of the examples use this. Should we fix these while we are making these improvements?

stc1988:
Thank you for confirming. In the scope of this pull request, we fixed the following changes:
* remove identifiers from JSON with comments.
* remove trailing comma

phoddie:
> remove identifiers from JSON with comments

That's a good solution. (I vaguely remember that there is a "JSON with Comments" variant in use in some ecosystems. Perhaps in the future that will be a valid identifier for GitHub markdown.)

> remove trailing comma

Perfect.

Thank you! We'll get this merged.

mkellner:
I have merged this. It will be available in the next code-drop.

stc1988:
> That's a good solution. (I vaguely remember that there is a "JSON with Comments" variant in use in some ecosystems. > > > Perhaps in the future that will be a valid identifier for GitHub markdown.)

I recently remembered this comment, and on GitHub, if you use the `jsonc` identifier, it will display correctly even with comments like the one below.

* jsonc
```jsonc
"include": [
	/* other includes here */
	"$(MODULES)/drivers/destm32s/manifest.json",
],
```

* json
```json
"include": [
	/* other includes here */
	"$(MODULES)/drivers/destm32s/manifest.json",
],
```

phoddie:
Thanks for remembering this! I've updated the markdown to use "jsonc" where appropriate. That should be available shortly.

Issue #1359, opened 2024-06-19
Issue with mcrun Command Freezing on M5Stack Core2


lamfan:
I am encountering an issue while trying to run mcrun to install a mod on the M5Stack Core2. The mcconfig command works fine, but mcrun freezes during the installation process.

### Steps to Reproduce

Successfully run the following command:
mcconfig -d -m -p esp32/m5stack_core -t build ./stackchan/manifest_local.json

Attempt to run the following command:
mcrun -d -m -p esp32/m5stack_core2 ./mods/look_around/manifest.json

### Environment

ESP-IDF version: esp-idf-v5.2.2
Operating System: macOS

phoddie:
Thank you for the report. 

In your "steps" to reproduce, the command line for the host is incorrect. It is only building (`-t build`) but not installing the host. Also, the platform is set to `esp32/m5stack_core` instead of `esp32/m5stack_core2`. Please try:

```
mcconfig -d -m -p esp32/m5stack_core2 ./stackchan/manifest_local.json
```

I don't have an M5Stack Core2. It is powered by an ESP32, the same as Moddable Two. I did a test of a host and mod on Moddable Two and it was successful.

lamfan:
@phoddie I forget I did run deploy after build

```
mcconfig -d -m -p esp32/m5stack_core2 -t build ./stackchan/manifest_local.json
mcconfig -d -m -p esp32/m5stack_core2 -t deploy ./stackchan/manifest_local.json
```

so I try your code 

```
mcconfig -d -m -p esp32/m5stack_core2 ./stackchan/manifest_local.json
```

after build it and I run 

```
mcrun -d -m -p esp32/m5stack_core2 ./mods/look_around/manifest.json
```

but it still freeze at Installing mod.
really don't understand what's wrong with it?

stc1988:
Hi @lamfan , Sharing capture of your terminal when you run mcrun will help us to understand the situation.

lamfan:
After running mcrun, it freeze at Installing mod. 
The M5Stack core2 did reload the screen, but nothing change....
<img width="761" alt="image" src="https://github.com/Moddable-OpenSource/moddable/assets/1715633/fd42bb3c-4871-4255-bb57-5fafd75b20e6">

phoddie:
[...]

PR #1388, opened 2024-08-10
change DEBUGGER_SPEED of m5stack_core2 (#1359)

stc1988:
I don’t know the end of this issue, but other users also encountered the same issue of not being able to run mod, and this has been resolved through this change.


mkellner:

PR #1389, opened 2024-08-11
fix syntax error in code block

stc1988:
I found some syntax error in code block by using [textlint-rule-eslint](https://github.com/textlint-rule/textlint-rule-eslint) and this PR fix these errors.


phoddie:
Very nice. This seems like a benefit of your earlier work to properly tag all JavaScript code in the markdown documents.

mkellner:

PR #1391, opened 2024-08-15
Add valid ID 32 for FT6206

washishi:
Added for M5Stack/CoreS3SE to return vendor ID32


washishi:
CoreS3 SE is Core S3 without cameras etc., so there is no problem specifying the build target as "esp32/m5stack_cores3"

phoddie:
Thanks for the PR. We'll get this merged

mkellner:

PR #1397, opened 2024-08-31
Fix for UUID() exception when time is defined and not using net stack

cmidgley:
If `Date().getTime()` indicates it has been set to recent time, `UUID()` will use `UUIDv1` which uses the Mac address (`Net.get("MAC")`) as part of the id.   However, if the network stack is not in use this results in an exception since it attempts to dereference `undefined`. 

This PR includes a test for `Net.get("MAC")` returning undefined and falling back on the same random number generator that `UUIDv4` uses.  

Not included in this PR is a possible improvement to `UUIDv4` as it could use the same logic (use Mac if it exists, else fallback on random) to improve the quality of the generated id when time is not defined but the network stack is available.    A simple copy/paste of the same line into the v4 version would add this (if interested).

### DIFF
--- a/modules/base/uuid/uuid.js
+++ b/modules/base/uuid/uuid.js
@@ -23,7 +23,7 @@
 	uuid
 */
 
-// https://tools.ietf.org/html/rfc4122
+// Compliant with UUIDv1 and UUIDv4 for RFC 9562 (supercedes RFC 5122) https://datatracker.ietf.org/doc/html/rfc9562 
 
 import Net from "net";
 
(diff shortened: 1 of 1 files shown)


phoddie:
The [RFC](https://datatracker.ietf.org/doc/html/rfc9562) makes this complicated. If I understand it correctly, using random bytes in place of the MAC address means it is no longer a v4 UUID, but custom (v8).

So... maybe we need a `UUIDv8` function. Since it is custom, it can use the MAC address when available and random digits otherwise. But.... please check out the RFC to confirm my reading. I could easily have misunderstood.

Also, in reading the code it looks like `hex()` should force `value` to an integer before calling `toString()`, Maybe this

```js
value = (value | 0).toString(16);
```

cmidgley:
Upon reading up on RFC 9562, I see that the prior UUID code was never in compliance.  Part of the issue was that `value` was not being cleaned up as an integer before hex conversion (resulting in a weird floating point hex number in the output), but also several issues with bit field sizes, incorrect field value settings (var/ver) and untruncated hex conversions.

To your point about a v1 UUID and random generated mac address, it is allowed as long as you set the multicast bit:

[RFC 9562 section 5.1](https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-1):

> The node field consists of an IEEE 802 MAC address, usually the host address or a randomly derived value per Sections [6.9](https://datatracker.ietf.org/doc/html/rfc9562#unguessability) and [6.10](https://datatracker.ietf.org/doc/html/rfc9562#unidentifiable)

[RFC 9562 section 6.10](https://datatracker.ietf.org/doc/html/rfc9562#name-uuids-that-do-not-identify-):

> Implementations MAY elect to obtain a 48-bit cryptographic-quality random number as per [Section 6.9](https://datatracker.ietf.org/doc/html/rfc9562#unguessability) to use as the Node ID. After generating the 48-bit fully randomized node value, implementations MUST set the least significant bit of the first octet of the Node ID to 1.

I have updated the code to be compliant with the RFC for a v1 (with Mac or randomly generated with broadcast set) and for v4, and have run these through several online UUID checkers/validators (v1 with and without mac, and v4), and they now all pass (including verifying ver and var).

P.S. If interested, this is the validation I use for UUID in my node/XS compatibility unit tests:

```js
        expect(/^[0-9a-f]{8}-[0-9a-f]{4}-(1|4)[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(id)).toBe(true);
```

phoddie:
Thanks for the detailed work to get this in compliance with the RFC. Much appreciated.  We'll get this merged.

PR #1401, opened 2024-09-02
add function define for AudioOut.Volume

stc1988:
It handles cases where [AudioOut.Volume](https://github.com/Moddable-OpenSource/moddable/blob/24e3e54fd3b66379c1f91ad92969c99f941ff4cf/documentation/pins/audioout.md#changing-volume) is used as follows.

```
audio.enqueue(0, AudioOut.Volume, 128);
```


stc1988:
I cofirmed this PR merged. Thank you.

PR #1413, opened 2024-09-25
Use CONFIG_ESP_CONSOLE_UART_BAUDRATE for instrumented ESP32 builds

cmidgley:
Update to all ESP32 `main.c` to use `CONFIG_ESP_CONSOLE_UART_BAUDRATE` instead of the hard-coded 115200 for instrumented builds (brief discussion on #1409).


mkellner:
This has been committed and will be available in the next drop.

PR #1420, opened 2024-10-05
fix: change `Headers` type to use `string` instead of `String` to resolve type error

meganetaaan:
The current type declaration used `String`, which caused an incompatibility error (TS2322) when assigning `Headers` to `Map<string, string | string[]>`. Changed to use `string` as it is the appropriate primitive type.

### DIFF
--- a/typings/headers.d.ts
+++ b/typings/headers.d.ts
@@ -19,7 +19,7 @@
 */
 
 declare module "headers" {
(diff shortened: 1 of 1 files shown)


phoddie:
This looks fine. We'll get it merged.

mkellner:

PR #1422, opened 2024-10-14
WiFi typings adjustment for API compatibility

cmidgley:
Small adjustment to the `WiFi` typings.  Before they used two types to define `mode` that were inconsistent with the JS API.  This PR removes those and defines the `WiFi.Mode.xxx` statics.

This is a breaking change due to the removal of the old types.  If it is preferred to be non-breaking, a possible solution might be to include them with JSdoc `@deprecated` tag, such as:

```js
  /**
   * @deprecated Use 'Mode' instead
   */
  export type StationMode = 1;
  /**
   * @deprecated Use 'Mode' instead
   */
  export type AccessPointMode = 2;
````


phoddie:
Thanks for updating this to sync with the change to make the constants visible in the API

The change is probably fine. Tagging @stc1988 and @meganetaaan in case this will be unacceptably disruptive for them.

Note that `3` is also a valid value, so you probably want to update the `WiFiMode` values.

> `// 3 reserved for station & accessPoint (WiFi.mode.station | WiFi.mode.accessPoint)`

cmidgley:
Added 3 to the `WiFiMode` type to allow `WiFi.Mode.station | WiFi.Mode.accessPoint`.

mkellner:

PR #1424, opened 2024-10-15
fix the type definition of the read function in AudionIn.

stc1988:
I noticed that the `read` function takes [a second and third argument.](https://github.com/stc1988/moddable/blob/public/modules/pins/audioin/esp32/audioin.c#L219-L230) and the type of return value is `ArrayBuffer`. 

I will make this correction in this PR. 
cc: @meganetaaan


mkellner:

PR #1427, opened 2024-10-19
syntax highlighting for code block and fix syntax error in code block

stc1988:
I haven’t checked all the documents again, but I found some new code blocks that need to be fixed.


phoddie:
Those all look great. Thank you!

mkellner:

Issue #1428, opened 2024-10-21
`embedded:io/socket/listener` locks up Win simulator


cmidgley:
**Build environment:** Windows
**Moddable SDK version:** 5.1.0
**Target device:** Win simulator.

**Description**
I'm trying to get `embedded:network/http/server` to work on `win` simulator, but it locks up (does not respond) the simulator process and requires it to be hard terminated.  The examples for `listener/httpserver` and `listener/listen` exhibit the failure, indicating the bug is likely related to `listener`.

Both examples work on `esp32`, but for `httpserver` a trivial bug in `examples/io/tcp/websocket/WebSocket.js` on line 51 needs to be fixed to not throw an error (doesn't affect the crash issue).  Change`if (href)` to `else if (href)`, as the prior check for `object` has identified this is an attachment and there is no `href`. 

**Steps to Reproduce**
1. Run `examples/io/listener/listen` or `examples/io/listener/httpserver` on windows simulator

phoddie:
Last I checked, the unit tests for Listener passed on Windows. We'll take another look.

phoddie:
The listener on Windows has a bug where it would end up as a blocking socket, rather than non-blocking. This led to the stalls you see. There are two places to fix that., Line 176:

```c
  			ioctlsocket(tcp->skt, FIONBIO, &nonBlocking);
```

And line 577:

```c
		ioctlsocket(listener->skt, FIONBIO, &nonBlocking);
```

There is also a smaller issue – a small memory leak if an exception occurs in the constructor – which will fixed when the above changes are integrated into the SDK.

cmidgley:
Patch applied and fix confirmed.  Thank you!

PR #1429, opened 2024-10-21
Typing improvements for `io` and fixes for `websocketclient`, and a few others

cmidgley:
This is the work I've done to date with `io` focused on making `websocketclient` to be reliable and strongly typed with TypeScript.  It's a fairly substantial set of changes, mostly in the `.d.ts` files but some implementation (especially `websocketclient`).  

Type changes:

- New types for `embedded:io/socket/tcp/tls`, `embedded:network/http/server`, `embedded:network/http/server/options/webpage`, `embedded:network/https/server/options/websocket`, and `embedded:network/websocket/client`
- Improvements to types `embedded:io/socket/listener`, `embedded:io/socket/tcp`,  `embedded:io/socket/udp`, `embedded:network/dns/resolver/udp`, `embedded:network/http/client`, `embedded:provider/builtin`
- Added `Device` (for global variable `device`) into global types to eliminate requirement for import
- Extend the global namespace for `Device` for when imports of extended modules are used (support for `embedded:network/http/server`, `embedded:network/websocket/client`)

Implementation changes:

- Bug fixes/improvements for `websocketclient`:
  - Implement error property via `onError(error: string)` (new feature)
  - Fix issue where control message payloads were not being decoded with the mask (data corruption)
  - Fix an issue where masks were used on both client and server (spec requires masking on client to server only)
  - Fixed issue where fragmented reads used an improper mask (data corruption)
- Fix a bug in `websocket` where using attached sockets would fail (reported in #1428)
- Add common HTTP status codes to `httpserver` on the `HTTP/1.1 <status> <code>` header
- Added options feature to `embedded:network/http/server/options/webpage` to support `.contentType` and `.status`. 

I typed everything based on the current implementation and not `ECMA-419` (many modules are inconsistent with the published 2nd edition of the spec).


examples/io/tcp/websocket/WebSocket.js
phoddie, line 1: I'm not sure about this change. It must fix a problem you see. But, it also would require the caller to set `host`, `port`, and `path` in the options object, which isn't what we want.

typings/embedded_network/websocket/client.d.ts
```diff
@@ -0,0 +1,56 @@
...
+			host?: string;
+			port?: number;
+			socket: TCPDevice | TLSDevice;
+		}) & {
+			protocol?: string;
+			headers?: Map<string, string | string[]>;
+			dns?: DNSUDPDevice;
+			onReadable?: (this: WebSocketClient, count: number, options: WebSocketClientReadableOptions) => void;
+			onWritable?: (this: WebSocketClient, count: number) => void;
+			onControl?: (this: WebSocketClient, opcode: WebSocketClientOpcode, buffer: Uint8Array) => void; // should this be ArrayBuffer?
```
phoddie: Yes, `ArrayBuffer` is correct as `onControl` is [specified](https://419.ecma-international.org/#-24-websocket-client-class-pattern-data-format).

examples/io/tcp/websocket/WebSocket.js
cmidgley: Good catch - on review I see I misunderstood what it was doing. I'll fix...

typings/embedded_network/websocket/client.d.ts
cmidgley: Oops.  That was an old comment that I forgot to remove.

cmidgley:
Fixes are in place.  Let me know what else you find..

typings/embedded_io/socket/tcp/tls.d.ts
```diff
@@ -21,12 +21,5 @@ declare module "embedded:io/socket/tcp/tls" {
...
 
-	export default class TLSSocket {
-		constructor(options: TLSOptions);
-		close(): undefined;
-		read(count: number | ArrayBufferLike): undefined | ArrayBufferLike;
-		write(buffer: ArrayBufferLike): number;
-		set format(format: string);
-		get format(): string;
-	}
+	export default class TLSSocket extends TCP {}
```
phoddie: I'm no TypeScript expert, but don't you need to define that the TLSSocket constructor uses `TLSOptions` instead of `TCPOptions` for the options object argument?

examples/io/listener/httpserver/httpserver.js
phoddie, line 1: Thank you for filling this in. The change is good. I don't want even more copies of the HTTP status code map in the repo. I'll make a module for those so it can be shared. But that doesn't need to hold this up.

typings/embedded_io/socket/udp.d.ts
phoddie, line 1: 419 UDP only specifies `udp.read()` - no argument and returns ArrayBuffer augmented with `port` and `address`. It does not specify multicast (though that's definitely on the roadmap)

typings/embedded_network/http/client.d.ts
```diff
@@ -20,12 +20,15 @@
...
-  export interface ClientOptions {
-    socket: any
-    port?:number
-    host: string
-    dns: any
+  export interface HTTPClientOptions {
+    socket: TCPDevice | TLSDevice;
+    port?: number;
+    host: string;
+    dns: DNSUDPDevice;
```
phoddie, line 31: There's no requirement that this is a DNS using UDP. It could be another resolution transport. But, for the moment this is probably OK?

cmidgley: I'd say we add a generic DNS type when we have more, and change ClientOptions to use that (or just DNSUDPDevice | DNSOtherDevice here).  Either way, it won't be a breaking change as it is just expanding the type.  But using 'any' (or 'undefined') isn't a good alternative as it isn't strongly typed.

cmidgley:
Merged latest code and resolved conflicts.  Reviewed `io` changes since initial PR and applied typing changes as needed.  Removed some no-longer-needed code changes as they are now in base.  Applied changes based on review comments from Peter.  This has been working great for me for the last seven months (on an older version of Moddable) and compiles/passes all my tests with latest code.

Ready for review and possible merging.

phoddie:
Finally coming back around to this.... there's so much here that it is a little challenging to review. I'm not asking that you split it up, but it is a bit of an adventure to understand and review.

That said, a lot of it looks good and we should get it merged. The `*-device` stuff is a bit confusing. I think I'm starting to understand it though.

cmidgley:
Closed as no longer aligns with latest typings work.

PR #1444, opened 2024-12-26
fix(build): skip manifest build var if already set

### LINKED ISSUE #1443: Build environment variables overwritten by nested manifests
**Build environment:** macOS
**Moddable SDK version:** 5.3..0
**Target device:**  NodeMCU ESP8266

**Description**
When targeting the ESP8266, the manifest for servo support includes a build environment variable for `ARDUINO_ESP8266` that is expected to be in `$(USERHOME)/esp/esp8266-2.3.0/`. This conflicts with the conventional setup location for xs-dev: `$(USERHOME)/.local/share/esp/esp8266-2.3.0/` (or `$(USERHOME)/xs-dev/esp/esp8266-2.3.0/` on Windows). Setting the `ARDUINO_ESP8266` build variable in my project manifest fails to fix the issue due to the [`parseBuild` method](https://github.com/Moddable-OpenSource/moddable/blob/public/tools/mcmanifest.js#L2171) in `mcmanifest.js` setting the environment variable, whether or not it has been set before. 

**Steps to Reproduce**
1. `xs-dev setup --device esp8266`
2. `xs-dev run --example pins/servo --device esp8266/nodemcu`
3. See error: `### Error: '$(ARDUINO_ESP8266)/libraries/Servo/src/esp8266/*': directory not found!`

**Expected behavior**
Setting the build environment variable should resolve the module paths correctly. 

**Other information**
Fixing this might resolve in breaking changes if other parts of the build system rely on this behavior,

HipsterBrown:
Fixes #1443

### DIFF
--- a/tools/mcbundle.js
+++ b/tools/mcbundle.js
@@ -103,6 +103,8 @@ export default class extends TOOL {
 		let properties = platform.build;
 		if (properties) {
 			for (let name in properties) {
+				if (this.environment[name])
+					continue;
 				let value = properties[name];
 				if (typeof value == "string")
 					this.environment[name] = this.resolveVariable(value);
--- a/tools/mcmanifest.js
+++ b/tools/mcmanifest.js
@@ -2172,6 +2172,8 @@ export class Tool extends TOOL {
 		let properties = platform.build;
 		if (properties) {
 			for (let name in properties) {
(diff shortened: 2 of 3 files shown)


phoddie:
This was solved in a different way (details in #1443).

PR #1446, opened 2024-12-30
fix(mcpack): check manifest include value before resolving

HipsterBrown:
While trying to build a project targeting the Raspberry Pi Pico with `mcpack`, I ran into an error:
```
### TypeError: call: not a function
```

All other platform targets were working, so I started exploring the cause.

After digging around the source code for `mcpack`, I narrowed down the issue to how the tool resolves modules from included manifest.json files. This led to the [arducam module](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/io/imagein/arducam_hm01b0/manifest.json#L6-L22) for the ECMA-419 image/in implementation, which uses a git include for the Pico. 

When comparing `mcpack` with the `mcmanifest` tool, I noticed the latter [has support for git sources](https://github.com/Moddable-OpenSource/moddable/blob/public/tools/mcmanifest.js#L1913) that is missing from `mcpack`. 

This PR fixes the current error by checking the type of the `include` field's value before attempting to resolve the expected string path to the included module. 

----

I'm not sure if git support should be added to mcpack if the evaluated manifest from mcpack still goes through mcconfig (a subclass of mcmanifest).

### DIFF
--- a/tools/mcpack.js
+++ b/tools/mcpack.js
@@ -1043,10 +1043,12 @@ export default class extends TOOL {
 	}
 	includeManifest(include, from, directory) {
 		this.currentDirectory = directory;
-		let path = this.resolveFilePath(this.resolveVariable(include));
-		if (!path)
-			throw new Error("'" + include + "': manifest not found!");
-		this.parseManifest(path, from);
+		if ("string" == typeof include) {
+			let path = this.resolveFilePath(this.resolveVariable(include));
+			if (!path)
+				throw new Error("'" + include + "': manifest not found!");
+			this.parseManifest(path, from);
+		}
 	}
 	mapBuiltins(builtins) {
 		const map = new Map;
--- a/tools/mcpack.js
+++ b/tools/mcpack.js
@@ -38,35 +38,35 @@ const fetchGlobal = {
 	snippet: `
 import { fetch } from "fetch";
 globalThis.fetch = fetch;
-`	
+`
 };
 const headersGlobal = {
 	include: "$(MODDABLE)/modules/data/headers/manifest.json",
 	snippet: `
 import Headers from "headers";
 globalThis.Headers = Headers;
-`	
+`
 };
 const structuredCloneGlobal = {
 	include: "$(MODDABLE)/modules/base/structuredClone/manifest.json",
 	snippet: `
 import structuredClone from "structuredClone";
 globalThis.structuredClone = structuredClone;
-`	
+`
 };
 const textDecoderGlobal = {
 	include: "$(MODDABLE)/modules/data/text/decoder/manifest.json",
 	snippet: `
 import TextDecoder from "text/decoder";
 globalThis.TextDecoder = TextDecoder;
-`	
+`
 };
 const textEncoderGlobal = {
 	include: "$(MODDABLE)/modules/data/text/encoder/manifest.json",
 	snippet: `
 import TextEncoder from "text/encoder";
 globalThis.TextEncoder = TextEncoder;
-`	
+`
 };
 const timerGlobal = {
 	snippet: `
@@ -75,7 +75,7 @@ globalThis.clearImmediate = globalThis.clearInterval = globalThis.clearTimeout =
 globalThis.setImmediate = function(callback) { return Timer.set(callback) };
 globalThis.setInterval = function(callback, delay) { return Timer.repeat(callback, delay) };
 globalThis.setTimeout = function(callback, delay) { return Timer.set(callback, delay) };
-`	
+`
 };
 const urlGlobal = {
 	include: "$(MODDABLE)/modules/data/url/manifest.json",
@@ -83,7 +83,7 @@ const urlGlobal = {
 import { URL, URLSearchParams } from "url";
 globalThis.URL = URL;
 globalThis.URLSearchParams = URLSearchParams;
-`	
+`
 };
 const workerGlobal = {
 	include: "$(MODDABLE)/modules/base/worker/manifest.json",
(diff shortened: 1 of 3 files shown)


phoddie:
Interesting find. The problem is very real. `mcpack` isn't expecting to find a Git repository in the includes. I'm not 100% sure what it should do here. I think the right approach is to make sure that the include gets passed through to `mcconfig` to handle. Will take a look.

HipsterBrown:
Bumping this PR since I'm running into the issue again while working on some new starting guides for the xs-dev docs and testing the example projects on the Raspberry Pi Pico. 

I also ran into an issue related to the PICO_EXTRAS_DIR and PICO_EXAMPLES_DIR paths being set in the pico's device manifest.json, which throws an error when trying to build with `mcpack` but not `mcconfig`; this behavior is due to the `resolveSource` in `mcpack.js` catching and reporting thrown errors while the same method in `mcconfig` seems to silently swallow them. I resolved this by allowing environment variables to take precedence over variables defined in the manifest's `build` config. The other solution would be to just remove those build vars from the pico's manifest.json and use the environment variables defined by developer / xs-dev only.

phoddie:
Thanks for putting this back on the active list and the additional fixes.

mcpack exists to allow you to work in the npm way. The git support is kind of a parallel world. But, they can collide. mcpack eventually outputs a manifest that mcconfig runs (at least that's my memory at the moment...) but obviously the contents of any git repos are going to be opaque to mcpack. Maybe that's ok though.

HipsterBrown:
Yes, this patch basically passes through the include config for that resource to mcconfig since it is unlikely to be a `package.json` based dependency that would need to be scanned by mcpack. It's also the only include defined that way AFAICT, so it's not worth pursuing full git support in mcpack at the moment.

phoddie:
I think I'm caught up. The original PR fixes `mcpack` so it doesn't stumble over Git includes. That makes complete sense. We should land that ASAP.

The second part is less obviously correct to me. It gives priority to environment variables over values defined in the `BUILD` section of the manifest. That's a behavior change which could have unintended side effects. Consider that we commonly define `"MODULES"` in `"BUILD"` as a shortcut to `$MODDABLE/modules`. If `MODULES` creeps in from the environment, there will be chaos. That's unlikely to happen by surprise for `PICO_EXTRAS_DIR ` and `PICO_EXAMPLES_DIR`, but the mechanism is general.

I understand why xs-dev wants to override these particular environment variables. What I think we want is equivalent to what we often do in make files, to provide a default value if the environment doesn't already provide one:

```
PICO_ROOT ?= $(HOME)/pico
PICO_SDK_DIR ?= $(HOME)/pico/pico-sdk
PICO_EXTRAS_DIR ?= $(HOME)/pico/pico-extras
```

In that spirit, I suggest the following  to signal that the environment variable should take priority. It is odd for a JSON name but it reasonably clear what is intended.

```json
"build": {
	"PICO_EXTRAS_DIR ?=": "$(HOME)/pico/pico-extras",
	"PICO_EXAMPLES_DIR ?=": "$(HOME)/pico/pico-examples"
}
```

HipsterBrown:
Thanks for taking the time to dig into this. 

I'm on board with that updated syntax for BUILD variables if they can't be moved to a Makefile as conditionally set variables (speaking to the Pico-specific ones in particular). There should be a clear call out in the docs about this convention as well. 

I understand the hesitation to just let the current environment variables take priority, especially with seemingly common names like "MODULES" or "IO". I know other tools have overcome this by namespacing variables that should be included in a program: https://vite.dev/guide/env-and-mode 
That's a larger code change but could allow for more resilient sandboxing of what mcconfig, et. al take in an inputs from the environment. It could even start with the `MODDABLE_MODULES` taking priority over the unprefixed `MODULES` build var internally. Every other toolchain has a similar namespace strategy for environment variables (ESP, PICO, ZEPHYR, NRF) except for the core Moddable tooling.

phoddie:
> I'm on board with that updated syntax for BUILD variables

Great. Do you prefer to do that or should I? 

With regard to naming, `BUILD` variables have always been set by a manifest either to be used by other manifests (`IO` and `MODULES`) or to propagate to the makefiles (`USE_USB`). We've never had a case, before the Pico variables discussed above, where the manifest defines a variable that it intends to allow to be overridden by the environment. I think it is best to be explicit about that – hence a distinct syntax. That way nothing unintentionally seeps in.

Cases like `IO `and `MODULES` are internal and should never be overridden. Adding a namespace defeats their purpose of simply shortening a path (`MODULES` is more compact than `$MODDABLE/modules`, `MODDABLE_MODULES` is almost the same).

HipsterBrown:
I can give it a shot. 

Thanks for the additional context about the intended use of BUILD vars. I'll keep that in mind for this implementation and future features. 👍

tools/mcpack.js
```diff
@@ -1080,7 +1082,16 @@ export default class extends TOOL {
...
+				const cleanName = name.replaceAll(' ?=', '');
+				const isConditional = name !== cleanName;
+
+				if (isConditional) {
+					const shellValue = this.getenv(cleanName);
+					if (shellValue !== undefined && !(cleanName in this.environment)) {
+						value = shellValue;
+					}
+				}
+				else if (typeof value == "string") {
```
HipsterBrown, line 1094: Here is the core of the logic for supporting the ` ?=` suffix for build variables in the manifest. Apologies for the extraneous formatting changes that happened from my editor.

phoddie:
Because of the white space adventure, I merged this manually. I rearranging things a bit and fixed an edge case (" ?=" should only be recognized at the end of a name, not anywhere). This will land in our May release, which should be later this month.

Thanks for your patience to work through this report. The details are a little subtle. The improvements are real.

PR #1447, opened 2025-01-04
[Enhancement] Added configuration of camera parameter

kitazaki:
camera_config.xclk_freq_hz is hardcoded in camera.c as 24000000.
It needs to define xclk_freq_hz as 20000000 below on OV3660 camera device.

I tried sample camera programs as camera-server-jpeg and camera-server-motion-jpeg on XIAO ESP32S3 Sense and M5Atom S3R M12.

* XIAO ESP32S3 Sense
OV2640 camera device
https://www.seeedstudio.com/XIAO-ESP32S3-Sense-p-5639.html

* M5Atom S3R M12
OV3660 camera device
https://docs.m5stack.com/en/core/AtomS3R-M12

OV2640 works fine for me, but OV3660 does not work correctly.
OV3660 outputs corrupted JPEG data with xclk_freq_hz as 24000000.
corrupt

I checked esp_camera.h in ESP32-Camera repository and tried a parameter same as xclk_freq_hz.
It works fine with 20000000.
https://raw.githubusercontent.com/espressif/esp32-camera/4335c93ec462a379aa2a3d655b8834972d725190/driver/include/esp_camera.h

### DIFF
--- a/examples/io/imagein/camera/camera-server-jpeg/main.js
+++ b/examples/io/imagein/camera/camera-server-jpeg/main.js
@@ -18,11 +18,13 @@ import Camera from "embedded:x-io/imagein/camera";
 import Net from "net";
 
 let width = 176, height = 144;
+let xclock = 20000000;
 let frame;
 
 const camera = new Camera({
 	width,
 	height,
+	xclock,
 	imageType: "jpeg",
 	format: "buffer/disposable",
 	onReadable: () => {
--- a/modules/io/imagein/camera/esp32/camera.c
+++ b/modules/io/imagein/camera/esp32/camera.c
@@ -336,6 +336,7 @@ void xs_camera_constructor(xsMachine *the)
 	Camera camera;
 	int imageType = kCommodettoBitmapRGB565LE;
 	uint8_t isJPEG = 0;
+	uint32_t xclock = 24000000;
 
 	xsmcVars(1);
 
@@ -362,6 +363,11 @@ void xs_camera_constructor(xsMachine *the)
 		else
 			imageType = xsmcToInteger(xsVar(0));
 	}
+	if (xsmcHas(xsArg(0), xsID_xclock)) {
+                xsmcGet(xsVar(0), xsArg(0), xsID_xclock);
+                xclock = xsmcToInteger(xsVar(0));
+		camera_config.xclk_freq_hz = xclock;
+        }
 
 	camera = c_calloc(1, sizeof(CameraRecord));
 	if (!camera)


phoddie:
Thank you for trying out the camera module and sharing your findings. We haven't had a chance to test with the OV3660, so your results are quite helpful!

All of the hardware configuration parameters for the camera are currently defined only in the manifest.  For example, here is the manifest fragment for the Xiao Sense board:

```jsonc
		"camera": {
			"xclk": 10,
			"pclk": 13,
			"href": 47,
			"vsync": 38,
			"scl": 39,
			"sda": 40,
			"d0": 15,
			"d1": 17,
			"d2": 18,
			"d3": 16,
			"d4": 14,
			"d5": 12,
			"d6": 11,
			"d7": 48
		}
```

Since the `xclk_freq_hz` is also a hardware configuration property, it will be fixed for any camera. Therefore, I think it makes sense to also define it in the manifest, rather than in script:


```jsonc
		"camera": {
			"xclk": 10,
			"xclk_freq_hz": 20000000,
			"pclk": 13,
			...
```

Then in the `camera.c` implementation, add this [near the top](https://github.com/Moddable-OpenSource/moddable/blob/9198d0111edab82bf68bd4e4747add00df0c3039/modules/io/imagein/camera/esp32/camera.c#L46):

```c
#ifndef MODDEF_CAMERA_XCLK_FREQ_HZ
	#define MODDEF_CAMERA_XCLK_FREQ_HZ (24000000)
#endif
```

...and change [this line](https://github.com/Moddable-OpenSource/moddable/blob/9198d0111edab82bf68bd4e4747add00df0c3039/modules/io/imagein/camera/esp32/camera.c#L162):

```c
	.xclk_freq_hz = MODDEF_CAMERA_XCLK_FREQ_HZ,
```

If we want to make xclk_freq_hz configurable from script, then we should probably make **all** the hardware camera properties configurable. I'm not sure if is necessary right now, however.

What do you think?

kitazaki:
@phoddie 
Thank you for your response. It looks reasonable for me for instance.

In addition to your proposal, camera_config.jpeg_quality is also for configurable.

sensor_t config in ESP32-Camera repository is important for image quality, but not defined in camera.c
brightness, contrast, saturation, sharpness and denoise are parameters for image quality.
I think this is furthermore enhancement.

(sorry for the blog page in Japanese, but good information for the difference between OV2640 and OV3660)
https://intellectualcuriosity.hatenablog.com/entry/2024/03/21/215308

phoddie:
@kitazaki – Thanks for reviewing. I'll make the change to add `XCLK_FREQ_HZ` and `JPEG_QUALITY` options to the manifest.

> brightness, contrast, saturation, sharpness and denoise are parameters for image quality.
> I think this is furthermore enhancement.

Completely agreed. These are values that we will want to change at runtime, so they will require an API. @mkellner has started to investigate this already. My hope is to standardized the API as part of ECMA-419 (the current Camera API is already scheduled to be part of ECMA-419 3rd Edition).

> ...good information for the difference between OV2640 and OV3660)

Very nice piece. Thanks for sharing!

Issue #1450, opened 2025-01-13
The characteristic class in BLE has a property `properties`.


stc1988:
When I was checking the data contained in the characteristic, I found a property `properties`, which is of the Number type.
This is not documented, but is it the same as what the GATT service has?

phoddie:
Yes, I think you are correct. The object passed to the `onCharacteristic` callback contains a `properties` property which corresponds to the GATT Characteristic Properties. It looks like we don't use that in the Moddable SDK examples.

Would you like to update the documentation or would you prefer that I do that?

stc1988:
the GATT Characteristic Properties comma-separating property strings like `"read,write"`, but  properties in the characteristic is number `18`, `22`.

I think it’s probably a combination of flags, but I’d like you to update the documentation to include that explanation as well.

phoddie:
The properties are a comma separated list in the JSON service description. Here they are flags, as you surmise. The definition of values comes from the BLE standard. Here is the relevant section of the ESP-IDF header `ble_gatt.h` that defines them:

```c
/**
 * @defgroup ble_gatt_chr_properties Generic Attribute Profile (GATT) Characteristic Properties
 * @{
 */

/** Characteristic property: Broadcast. */
#define BLE_GATT_CHR_PROP_BROADCAST                     0x01

/** Characteristic property: Read. */
#define BLE_GATT_CHR_PROP_READ                          0x02

/** Characteristic property: Write Without Response. */
#define BLE_GATT_CHR_PROP_WRITE_NO_RSP                  0x04

/** Characteristic property: Write. */
#define BLE_GATT_CHR_PROP_WRITE                         0x08

/** Characteristic property: Notify. */
#define BLE_GATT_CHR_PROP_NOTIFY                        0x10

/** Characteristic property: Indicate. */
#define BLE_GATT_CHR_PROP_INDICATE                      0x20

/** Characteristic property: Authenticated Signed Write. */
#define BLE_GATT_CHR_PROP_AUTH_SIGN_WRITE               0x40

/** Characteristic property: Extended Properties. */
#define BLE_GATT_CHR_PROP_EXTENDED                      0x80
```

It appears that only READ, WRITE_NO_RSP, WRITE, NOTIFY, INDICATE and EXTENDED are currently used.

For the documentation, I suggest adding `properties` to the table in Class Characteristic > Properties:

| Name | Type | Description |
| --- | --- | :--- |
| `...` | `...` | ...
[...]

PR #1452, opened 2025-01-16
Split ble.d.ts and update the type definitions.

stc1988:
I'm trying to BLE client application using TypeScript, but I found that name of modules was not resolved to import.
So this PR split ble.d.ts into each modules and update some type definitions of these.

Off course, this includes update of #1450.


phoddie:
The change makes sense. We'll get this merged. Thank you.

mkellner:
This has been committed. It will be available in the next release.

PR #1455, opened 2025-01-22
Fix the type definition of Bytes.

stc1988:
I split the definition of ble in #1452  , but the definition of [Bytes](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/ble/btutils.js#L131-L163) does not match the implementation, so it will be corrected.


phoddie:
The change looks good. We'll integrate that. Thank you.

mkellner:
This will be available in the next source code update.

PR #1456, opened 2025-01-22
Split `m5atom_s3r` into `m5atom_s3r_cam` and `m5atom_s3r_m12`.

stc1988:
[M5AtomS3R](https://shop.m5stack.com/products/atoms3r-dev-kit?variant=45605332615425) is the PSRAM version of M5AtomS3.

In this PR, the device that has  camera will be defined as the following two.
* [m5atom_s3r_cam](https://shop.m5stack.com/products/atoms3r-camera-kit)
* [m5atom_s3r_m12](https://shop.m5stack.com/products/atoms3r-camera-kit-m12-version-ov3660)

Each is mostly just renamed, but in accordance with #1447, `xclk_freq_hz` is defined as `20000000` for `m5atom_s3r_m12`. 
cc: @kitazaki


phoddie:
Thanks – so many devices to keep track of!

mkellner:
Thank you.
This will be available in the next source code drop.

PR #1473, opened 2025-03-20
add typing  embedded:io/audio/in and embedded:io/audio/out

stc1988:
This PR adds typings for `embedded:io/audio/in` and `embedded:io/audio/out`

### DIFF
--- a/typings/embedded_io/audioin.d.ts
+++ b/typings/embedded_io/audioin.d.ts
@@ -1,0 +1,51 @@
+/*
+* Copyright (c) 2025 Satoshi Tanaka
+*
+*   This file is part of the Moddable SDK Tools.
+*
+*   The Moddable SDK Tools is free software: you can redistribute it and/or modify
+*   it under the terms of the GNU General Public License as published by
+*   the Free Software Foundation, either version 3 of the License, or
+*   (at your option) any later version.
+*
+*   The Moddable SDK Tools is distributed in the hope that it will be useful,
+*   but WITHOUT ANY WARRANTY; without even the implied warranty of
+*   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+*   GNU General Public License for more details.
+*
+*   You should have received a copy of the GNU General Public License
+*   along with the Moddable SDK Tools.  If not, see <http://www.gnu.org/licenses/>.
+*
+*/
+declare module "embedded:io/audio/in" {
+  import { Buffer } from "embedded:io/_common";
+
+  type AudioType = "LPCM";
+
+  interface AudioInOptions {
+    bitsPerSample?: 8 | 16;
+    channels?: 1 | 2;
+    sampleRate?: number;
+    audioType?: AudioType;
+    format?: "buffer";
+  }
+
+  class AudioIn {
+    constructor(options: AudioInOptions & {
+      onReadable?: (this: AudioIn, bytes: number) => void;
+    });
+    read(byteLength?: number): ArrayBuffer
+    read(buffer: Buffer): void;
+    start(): void;
+    stop(flush?: boolean): void;
+    close(): void;
+    readonly bitsPerSample: 8 | 16;
+    readonly channels: 1 | 2;
+    readonly sampleRate: number;
+    readonly audioType: AudioType;
+    get format(): "buffer"
+    set format(value: "buffer")
+  }
+
+  export { AudioIn as default };
+}
--- a/typings/embedded_io/audioin.d.ts
+++ b/typings/embedded_io/audioin.d.ts
@@ -28,16 +28,17 @@ declare module "embedded:io/audio/in" {
     sampleRate?: number;
     audioType?: AudioType;
     format?: "buffer";
+    target?: any;
   }
 
   class AudioIn {
     constructor(options: AudioInOptions & {
-      onReadable?: (this: AudioIn, bytes: number) => void;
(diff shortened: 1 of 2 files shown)


typings/embedded_io/audioin.d.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  class AudioIn {
+    constructor(options: AudioInOptions & {
+      onReadable?: (this: AudioIn, bytes: number) => void;
+    });
+    read(byteLength?: number): ArrayBuffer
+    read(buffer: Buffer): void;
+    start(): void;
+    stop(flush?: boolean): void;
```
phoddie: This should be `stop(options?: { flush?: boolean }): void`  like you have for `AudioOut`.

typings/embedded_io/audioin.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  type AudioType = "LPCM";
+
+  interface AudioInOptions {
+    bitsPerSample?: 8 | 16;
+    channels?: 1 | 2;
+    sampleRate?: number;
+    audioType?: AudioType;
+    format?: "buffer";
+  }
```
phoddie, line 32: All constructor options objects in the IO Class Pattern have an optional `target` property, inherited from the [Base Class Pattern constructor](https://ecmatc53.github.io/spec/web/spec.html#-8-base-class-pattern-constructor). It can be any JavaScript value, which I think is `any` in TypeScript?

typings/embedded_io/audioin.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  class AudioIn {
+    constructor(options: AudioInOptions & {
+      onReadable?: (this: AudioIn, bytes: number) => void;
+    });
+    read(byteLength?: number): ArrayBuffer
+    read(buffer: Buffer): void;
```
phoddie: Return type is `number` when passing a `Buffer` - see the [`read` method](https://ecmatc53.github.io/spec/web/spec.html#-9-io-class-pattern-read-method) of the IO Class Pattern.

typings/embedded_io/audioin.d.ts
```diff
@@ -0,0 +1,51 @@
...
+    bitsPerSample?: 8 | 16;
+    channels?: 1 | 2;
+    sampleRate?: number;
+    audioType?: AudioType;
+    format?: "buffer";
+  }
+
+  class AudioIn {
+    constructor(options: AudioInOptions & {
+      onReadable?: (this: AudioIn, bytes: number) => void;
```
phoddie: There is a second argument to the `onReadable` callback – the `sampleCount` – see onReadable in the [Audio Input specification](https://ecmatc53.github.io/spec/web/spec.html#-10-io-classes-audio-input-synchronous-io).

typings/embedded_io/audioout.d.ts
```diff
@@ -0,0 +1,49 @@
...
+    start(): void;
+    stop(options?: { flush?: boolean }): void;
+
+    readonly audioType: AudioType;
+    readonly bitsPerSample: 8 | 16;
+    readonly channels: 1 | 2;
+    readonly sampleRate: number;
+    get format(): "buffer"
+    set format(value: "buffer")
+    volume: number;
```
phoddie, line 48: If TypeScript allows expressing ranges, the `volume` property is defined to be between 0.0 and 1.0 inclusive.

typings/embedded_io/audioout.d.ts
```diff
@@ -0,0 +1,49 @@
...
+  import { Buffer } from "embedded:io/_common";
+
+  type AudioType = "LPCM";
+
+  interface AudioOutOptions {
+    bitsPerSample?: 8 | 16;
+    channels?: 1 | 2;
+    sampleRate?: number;
+    audioType?: AudioType;
+    format?: "buffer";
```
phoddie, line 30: This should include `target`, as with `AudioInOptions`.

phoddie (changes requested):
This looks really good. Thank you for taking the initiative!

I noted a few items that should be corrected. Please let me know if you have any questions about those.

Note that these typings are only for the synchronous Audio Input and Audio Output. That's practical for now, as that's what the Conversational AI work is using.

typings/embedded_io/audioout.d.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  class AudioOut {
+    constructor(options?: AudioOutOptions);
```
phoddie: This should include `onWritable(byteLength, sampleCount)` – see [Callbacks in Audio Output specification](https://ecmatc53.github.io/spec/web/spec.html#-10-io-classes-audio-output-synchronous-io).

phoddie (changes requested):

typings/embedded_io/audioout.d.ts
stc1988: May be  can’t specify a range with the type

stc1988:
Thank you for the review.I’ve made the changes according to the review. Please check.

> Note that these typings are only for the synchronous Audio Input and Audio Output. 

I see. The asynchronous pattern may need to be supported in other classes as well.
I think I’ll review it at another time.

phoddie:
Thanks for the quick update. The changes look good. We'll get this merged.

mkellner:
This will be available in the next code drop.

PR #1476, opened 2025-03-22
instructions option  for openai-stream

stc1988:
In OpenAI’s latest [announcement](https://openai.com/index/introducing-our-next-generation-audio-models/), a new audio model was introduced.
This PR adds an `instructions` option that can be used with the `gpt-4o-mini-tts` model for text-to-speech.

I also confirmed that the Realtime API allows specifying a model for transcription and that it works, but since Gemini didn’t have a similar option, I didn’t extend `chatAudioIO`.


phoddie:
I think the changes to the read.md are unrelated?

stc1988:
Sorry for the lack of explanation.
This fix is not directly related, but since it’s a trivial change, I added the commit to this PR.

mkellner:
This will be available in the next code drop.

PR #1482, opened 2025-04-05
Update the camera documentation following #1456

stc1988:
I fixed the omission in the documentation that occurred  in #1456.


mkellner:

PR #1483, opened 2025-04-05
rename  from typings/embedded_io/audio* to typings/embedded_io/audio/*

stc1988:
Sorry for the insufficient conrimination of #1473. I fixed the path of the typing files.


phoddie:
No problem. These kinds of mistakes are easy to make and, fortunately, easy to correct.

mkellner:

PR #1487, opened 2025-04-22
Fixing udp.d.ts

linfan68:
Found an Error on the .d.ts file (Buffer should be the 3rd argument), refer to different native c code:

https://github.com/Moddable-OpenSource/moddable/blob/2945c5d0f2d37e534a3d09729f25e576b0561676/modules/io/socket/win/udp.c#L226

https://github.com/Moddable-OpenSource/moddable/blob/2945c5d0f2d37e534a3d09729f25e576b0561676/modules/io/socket/lin/udp.c#L233

https://github.com/Moddable-OpenSource/moddable/blob/2945c5d0f2d37e534a3d09729f25e576b0561676/modules/io/socket/lwip/udp.c#L197

https://github.com/Moddable-OpenSource/moddable/blob/2945c5d0f2d37e534a3d09729f25e576b0561676/modules/io/socket/mac/udp.m#L237

### DIFF
--- a/typings/embedded_io/socket/udp.d.ts
+++ b/typings/embedded_io/socket/udp.d.ts
@@ -32,7 +32,7 @@ declare module "embedded:io/socket/udp" {
       multicast: string;
       timeToLive: number;
     }))
-    write(buffer: Buffer, address: string, port: number): void;
+    write(address: string, port: number, buffer: Buffer): void;
     read(): ArrayBuffer & {
       address: string;
       port: number;


phoddie:
Yikes!

You are correct that `udp.d.ts` is out of sync with the implementations. But.... the implementations are wrong! Checking the [standard](https://419.ecma-international.org/#-10-io-classes-udp-socket) for UDP, the `buffer` argument is first. That's consistent with the IO Class Pattern which always defines the first argument as the data to be written and leaves the remaining arguments, if any, to the particular class.

I think the implementations should be fixed. That will break some code. Best to do that sooner than later. For the transition, I can check for the first argument being a string and issue a specific exception about the needed change.

phoddie:
Closing as the implementation has been updated to match the 419 specification and type declaration.

PR #1495, opened 2025-05-06
Add escaping for private scoped modules when building TypeScript on non-windows hosts

cmidgley:
Fix for private-scoped modules where the `#` symbol wasn't being escaped correctly (the tsconfig dependency rule, on non-Windows platforms).


mkellner:

PR #1498, opened 2025-05-11
Added Atomic Echo Base device

kitazaki:
<A HREF="https://docs.m5stack.com/en/atom/Atomic%20Echo%20Base">Atomic Echo Base</A> is speaker and microphone base for M5 Atom series.

I added M5Atom Lite and M5Atom S3 with Atomic Echo Base target files.

m5atom_s3r_echo_base_1
m5atom_s3r_echo_base_2

This is a sample. (small talking stack-chan, Node-RED MCU Edition)
img_0483_720


mkellner:
This will be available in the next release.

PR #1503, opened 2025-06-06
m5atom s3 simulator

stc1988:
As chance to use the M5Atom S3 have increased, such as adding the M5Atom Echo Base target, I have added a simulator.

スクリーンショット 2025-06-07 0 15 43


phoddie:
Fun.

PR #1505, opened 2025-06-10
add Typing  for Digest

stc1988:
There are other classes in the crypt module, and many classes related to encryption, but I think `Digest` is the one that is most often used directly by applications, so this PR only covers `Digest`.


phoddie:
`Digest` makes perfect sense as a starting point. Thank you.

PR #1508, opened 2025-06-15
migrate M5StackCoreTouch to ECMA 419

stc1988:
In [fabf5e6](https://github.com/Moddable-OpenSource/moddable/commit/fabf5e66856717585099a976bbc26632922626cf), `M5StackCoreTouch` uses 419 driver, but virtualbutton doesn't work by this change.

This PR changes M5StackCoreTouch to conform to the ECMA-419 interface.


stc1988:
@phoddie 
I have confirmed that it also works with the new I²C in version 5.10.0. Could you confirm this PR?

mkellner:
This has been committed and will be available in the next code drop.

PR #1519, opened 2025-07-19
getCharacteristics of web-bluetooth can find all characteristics

stc1988:
I quickly started testing the new Web Bluetooth API, but I found an issue that `getCharacteristics` doesn’t work and I can’t retrieve all the characteristics.

This PR fixes the above issues.

### DIFF
--- a/modules/io/ble/web-bluetooth/web-bluetooth.js
+++ b/modules/io/ble/web-bluetooth/web-bluetooth.js
@@ -314,9 +314,14 @@ class BluetoothDevice {
 			return target.promise;
 		}
 		getCharacteristics(ids) {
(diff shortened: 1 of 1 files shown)


stc1988:
This is a separate matter from the topic, but I made additional modifications because I couldn’t retrieve the characteristic properties.

phoddie:
Thank you for trying this out so quickly. Very helpful.

The `new new` is clearly wrong.

The `getCharacteristics` change is good but can be more concise:

```js
ids = ids?.map(id => BluetoothUUID.getCharacteristic(id));
```

That same change needs to be applied to `getPrimaryService()` and `getDescriptors()`. We can do that after merging the PR.

mkellner:
This is now available in TOT.

PR #1522, opened 2025-08-03
Update Timer typings to use Timer IDs

stc1988:
Refactored the `Timer` type definitions to use numeric timer IDs instead of Timer instances. 
Although not included in this repository, I think it is also necessary to modify the following functions defined in `mcpack.d.ts` of  `@moddable/typings`   so that their return values are IDs.

```
declare function setInterval(handler: Function, timeout?: number): void;
declare function setTimeout(handler: Function, timeout?: number): void;
```

### DIFF
--- a/typings/timer.d.ts
+++ b/typings/timer.d.ts
@@ -19,7 +19,7 @@
 */
 
 declare module "timer" {
-  export type TimerCallback = (timer: Timer) => void;
+  export type TimerCallback = (id: number) => void;
 
   class Timer {
     private constructor()
@@ -27,20 +27,18 @@ declare module "timer" {
       callback: TimerCallback,
       interval?: number,
       repeat?: number
-    ) => Timer
+    ) => number
     static repeat: (
       callback: TimerCallback,
       interval: number,
-    ) => Timer
+    ) => number
     static schedule: (
-      timer: Timer,
+      id: number,
       interval?: number,
       repeat?: number
     ) => void;
-    static clear: (timer: Timer | undefined | null) => void;
+    static clear: (id: number | undefined | null) => void;
     static delay: (milliseconds: number) => void;
-
-    private brand: boolean;
   }
 
   export {Timer as default};


phoddie:
Thanks for the PR. I'm not sure we should make this change. Let me try to explain.

My guess is that you are proposing this change to match the behavior of `setInterval` and friends from the web. That is understandable, but the value returned by `Timer.set` and friends is really an object, not a number. Forcing the type to be a number means that developers could write code that fails. The following is a little silly, but would be considered valid by TypeScript, yet would fail:

```
const timers = new Float64Array(1);
timers[0] = setInterval(() => {debugger}, 1);
clearInterval(timers[0]);
```

There is another popular JavaScript runtime where `setInterval` and friends return an object: Node.js. You can check Node's `Timer` [docs for the details](https://nodejs.org/api/timers.html#settimeoutcallback-delay-args). The Moddable SDK typing for Timer are consistent with that approach.

The use of a number by `setInterval` and friends to represent a timer is historic and not without problems. Still, if you really want or need `setInterval` and friends to use numbers for timers, the right solution under the Moddable SDK would probably be to provide a separate implementation of those global functions that maintains a private global `Map` to convert between the timer instance and its corresponding number.

stc1988:
Sorry, I forgot to check what `Timer.set` actually returns.  I understand that Timer.set returns an object, but I have two points to confirm.

- The expression “timer’s ID” in the [documentation](https://github.com/Moddable-OpenSource/moddable/blob/1bea8ad83750c0455ebb83e4ee5243d84e0745e1/documentation/base/base.md#timer) evokes an interface similar to the Web, but wouldn’t it be more appropriate to state that it returns a timer object?
- I’m not sure whether `setInterval` in `mcpack` is compatible with the Web or Node.js, but I would like to add a type definition for its return value.

phoddie:
> The expression “timer’s ID” in the [documentation](https://github.com/Moddable-OpenSource/moddable/blob/1bea8ad83750c0455ebb83e4ee5243d84e0745e1/documentation/base/base.md#timer) evokes an interface similar to the Web, but wouldn’t it be more appropriate to state that it returns a timer object?

If I recall well, the Timer API returned a number when the documentation was written. I'll update the docs to avoid any confusion.

The documentation also says...

> Timer IDs are opaque that are only useful for passing to Timer functions

...which remains accurate. There are currently no methods on Timer instances beyond those inherited from `Object.prototype`. That answers your question about `setInterval()` in `mcpack`: the instance is opaque. It is not just a POJO, however; it is a Timer. That's why the  TypeScript declaration of Timer includes `private brand: boolean` – so that only true Timer instances may be passed to the Timer API. Maybe that means that `mcpack.d.ts` should be updated to replace `id: number` with `id: Timer`.

stc1988:
Thank you for the explanation, I understand now. I’ll close this PR as it’s unnecessary.

phoddie:
I think something like this is needed in mcpack.d.ts – changing the timer values from numbers to Timer instances.

```ts
declare function clearImmediate(id: Timer): void;
declare function clearInterval(id: Timer): void;
declare function clearTimeout(id: Timer): void;
declare function setImmediate(handler: Function): Timer;
declare function setInterval(handler: Function, timeout?: number): Timer;
declare function setTimeout(handler: Function, timeout?: number): Timer;
```

What do you think?

stc1988:
Got it, same here.

stc1988:
typings package has been updated to 5.10.0, but does it include the mcpack.d.ts update?

phoddie:
> , but does it include the mcpack.d.ts update?

No, it does not. I should have posted about this earlier.

When I tried the change to mcpack.d.ts locally, TypeScript generated all sorts of errors. I tried to find a solution but failed. Were you able to make the change and use it successfully with `mcpack`?

stc1988:
I was able to solve it with the following definition.

``` ts
interface Console {
	log(...log: (string | number | boolean)[]):void;
}

import type Timer from "timer";

declare global {
 const console: Console;
 const clearImmediate: (timer: Timer) => void;
 const clearInterval: (timer: Timer) => void;
 const clearTimeout: (timer: Timer) => void;
 const setImmediate: (handler: Function) => Timer;
 const setInterval: (handler: Function, timeout?: number) => Timer;
 const setTimeout: (handler: Function, timeout?: number) => Timer;
}
```

When I try to import Timer, it becomes a [top-level import and seems to be defined as a module](https://www.typescriptlang.org/docs/handbook/2/modules.html?utm_source=chatgpt.com#how-javascript-modules-are-defined), and it seems that in order to declare something as global in a modular file, we need to use `declare global`.

phoddie:
Hey, that works nicely. Thank you. It seems obvious now. ;) TypeScript is nice to use, but creating the declarations still feels mysterious to me.

We'll publish an update to @moddable/typings so that this is available before the next Moddable SDK release.

stc1988:
Version 5.10.1 works as expexted. Thank you.

Issue #1527, opened 2025-09-03
[Feature request] Update to the new model `gpt-realtime`


stc1988:
New speech-to-speech model `gpt-realtime` is now [generally available](https://openai.com/index/introducing-gpt-realtime/).

My understanding is that this includes the following advantages:
- Higher performance and lower cost of use
- two new voices, Cedar and Marin
- Remote MCP server support
- Image Input

I tried using this new model, but there seem to be changes in the message structure, which will require some work.

stc1988:
[As a result of trial and error,](https://github.com/Moddable-OpenSource/moddable/compare/public...stc1988:moddable:feature/gpt-realtime) I was able to have conversations with the conversational AI app with new voices. I am currently checking the tool function.

stc1988:
Since we hadn’t been able to continue the conversation after sending the functionResult even before this change, I applied a fix and confirmed it’s working.
For now, I’ll create a PR with the fixes up to this point.

PR #1529, opened 2025-09-13
Update to the new model gpt-realtime (#1527)

stc1988:
This PR adds support for the new openAIRealtime model `gpt-realtime`(#1527).

I confirmed that conversationalAI and chataudioio-tools are working.


phoddie:
Very interesting. I'll take a look.

stc1988:
I think [Moddable 6.1.0](https://github.com/Moddable-OpenSource/moddable/releases/tag/6.1.0) did this update.

PR #1531, opened 2025-09-22
Add MCP23017 expander module to manifest

stc1988:
Updated `manifest.json` to include the MCP23017 expander module and preload it. This allows the example to use the MCP23017 I/O expander functionality.

### DIFF
--- a/examples/io/i2c/expander/main.js
+++ b/examples/io/i2c/expander/main.js
@@ -12,7 +12,7 @@
  *
  */
 
-import Expander from "embedded:io/provider/MCP23017";
+import Expander from "embedded:io/provider/MCP23X17";
 
 const expander = new Expander({
 	i2c: device.I2C.default,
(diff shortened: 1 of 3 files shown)


phoddie:
Yes, this is out of date in a few ways. Thank you for pointing that out.

Our usual style these days is to have a manifest for the module, so projects can include that directly. Here that would mean adding a manifest at `$MODDABLE/modules/io/expander/manifest.json`:

```json
{
	"modules": {
		"embedded:io/provider/MCP23X17": "./expander"
	},
	"preload": [
		"embedded:io/provider/MCP23X17"
	]
}
```

The module specifier used in the example is also slightly incorrect. It should be `embedded:io/provider/MCP23X17` to match the [specification](https://github.com/EcmaTC53/spec/blob/master/docs/Hardware%20Components/providers/Digital-Microchip-MCP23X17.md) referenced by [ECMA TR/109](https://ecma-international.org/publications-and-standards/technical-reports/ecma-tr-109/).

stc1988:
@phoddie 
Thank you for reviewing. I updated the source code.

mkellner:
This will be available in the next release.

PR #1534, opened 2025-09-23
Fixed mbedtls include path

rmontrosecbw:
The mbedtls include path in Espressif SDK version 5.5.x is $(IDF_PATH)/components/mbedtls/mbedtls/include/. This is correct in make.esp32.mk but was left out of nmake.esp32.mk


phoddie:
Makes sense. Thank you. We'll get this merged.

mkellner:
This has been committed and will be available in the next code drop.

PR #1539, opened 2025-10-17
Added FreeRTOS and ESP32 Core settings to worker configuration

rmontrosecbw:
Added the ability to set FreeRTOS task priority setting for workers, and for ESP32 devices, assigning the worker task to a specific CPU core.
```js
let aWorker = new Worker("simpleworker", {
	static: 8192,		// bytes
	stack: 64,			// JavaScript stack in 16 byte slots
	heap: {				// Values in 16 byte slots
		initial: 64,
		incremental: 32
	},
	nativeStack: 8192,	// host's native "C" stack, in bytes
	priority: 3,		// For FreeRTOS, sets the Worker Task priority
	coreId: 1			// For ESP32, set the execution core. 0 or 1
});
```
Enhanced `worker.md` documentation to reflect the changes.

### DIFF
--- a/documentation/base/worker.md
+++ b/documentation/base/worker.md
@@ -49,7 +49,7 @@ let aWorker = new Worker("simpleworker", {
 	},
 	nativeStack: 8192,	// host's native "C" stack, in bytes
 	priority: 3,		// For FreeRTOS, sets the Worker Task priority
-	coreId: 1			// For ESP32, set the execution core. 0 or 1
+	core: 1				// For ESP32, set the execution core. 0 or 1
 });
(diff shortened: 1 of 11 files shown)


phoddie:
The priority and core additions are welcome. They may eventually be useful on other platforms.

A couple of items to correct:

- No need to extend the native `xsCreation` structure (and definitely not across every platform XS runs on). It looks like the new properties could be local variables in `workerConstructor()`. 
- `coreId` should be `core` – the `Id` doesn't add information.

mkellner:

PR #1540, opened 2025-10-17
Worker subclass Example

rmontrosecbw:
Showing how to create a class derived from the Worker class. You need a separate coreworker.js function for the task.


mkellner:

PR #1541, opened 2025-10-26
Wiznet w5500 ethernet support

rmontrosecbw:
You identified a problem with deprecated ESP Ethernet calls, so I got a few WT32-ETH01 boards to test the changes. They didn't work with the changes, and that led me down a rabbit hole to get them working. I don't know if Espressif changed how their stack works, but with these boards (probably all internal PHY devices) and the Espressif examples, you need to initialize netif and the event loop before creating netif objects (ethernet.c used first to make the network objects). While tracking down the problem, I added many debugging statements and gated them with the Moddable define ethernet_debug. We need to switch between static and DHCP, so I've added methods to support this and to report the IPv6 address (we will be supporting that). There are several additional MODDEF flags to support the WT32-ETH01 boards.
I didn't think this would require as many changes but it now allows better monitoring of the ESP32 Ethernet behavior.


phoddie:
Thanks so much for the diligent work. And for your patience on. We've been [busy](https://github.com/Moddable-OpenSource/moddable/releases/tag/6.0.0). I expect we'll get through a review on this in the next week and merged from there.

phoddie:
Hey – so I tried this with a (newly acquired) WT32-ETH01 board using the `esp32/wt32_eth01` board target. It builds and runs, but it fails to find the ethernet hardware:

```
~moddable/modules/network/ethernet/esp32/ethernet.c (491) # Break: Error: Ethernet.c No PHY found on MDIO bus! Check wiring and clock. (in start)!
```

In reading the code I noticed this just above:

```c
#ifdef MODDEF_ETHERNET_INTERNAL_PHY_GPIO0_CLOCK_INPUT
	// Configure GPIO0 as input to prevent conflict with external 50MHz oscillator (RMII_CLK)
	// This is critical for WT32-ETH01 and similar boards with external clock source
```

`MODDEF_ETHERNET_INTERNAL_PHY_GPIO0_CLOCK_INPUT` isn't defined. As a test, I defined it but that didn't change the behavior.

I also tried uncommenting this block:

```c
	// Explicit clock configuration (uncomment if sdkconfig settings aren't working)
	// esp32_mac_config.clock_config.rmii.clock_mode = EMAC_CLK_EXT_IN;
	// esp32_mac_config.clock_config.rmii.clock_gpio = EMAC_CLK_IN_GPIO;
```

From what I understand, you would expect this board to work with the changes. Can you confirm that the code is as expected. If so, perhaps the hardware is bad or a different board revision?  For what it is worth, here's [the board](https://www.amazon.com/dp/B0BYXJTWPB) I'm using.

Thanks!

phoddie:
FWIW – I couldn't get the WT32-ETH01 working. I got an Espressif Ethernet Kit instead. With a couple of small configuration changes, that worked perfectly. Cool. I'm going to assume that there are variations the WT32-ETH01 in the wild and with some patience and better luck I could have gotten it working.

We've merged this PR. I made some small adjustments for consistency with the rest of the Moddable SDK. I also updated the documentation with your new APIs. ;)

The changes will land in the next Moddable SDK release. Thanks again!

mkellner:

rmontrosecbw:
I had two different WT32-ETH01 boards (one with the pins on the bottom and one, like your's, with the pins on the top and the code I pushed worked with both of them. Though there are a lot of those clones out there and some may be different. I ordered the board you pointed to and will be curious if I can find out what's different about it.

PR #1549, opened 2025-12-21
Add typings for neopixel and neostrand

stc1988:


phoddie:
Thank you for preparing these type definitions. They are well done. 

I noticed two small details.

First, the Neopixels constructor uses `optios` instead of `options`.

Second, the Neopixels instance is also a host buffer. This allows for script optimizations. The details are explained in [NeoPixels, Moddable style](https://www.moddable.com/blog/neopixels/) in the section "Direct pixel access in JavaScript."

```js
let pixels = new Uint32Array(np);
pixels[0] = white;
pixels[1] = ~pixels[0];
```

I think you just need to change the Neopixels type to extend `HostBuffer` from `xs.d.ts`. But, if that gets complicated, you could skip it for now.

stc1988:
Thank you for reviewing in detail. I’ve made the corrections based on your comments.

mkellner:

PR #1552, opened 2025-12-31
Feature/cores3 mic

meganetaaan:
This is a draft PR to add microphone support for M5Stack CoreS3

- Add microphone (ES7210 audio ADC) configuration for M5Stack CoreS3 setup-target sequence
- Extend AudioIn class to support stereo input, I2S Philips format, and MCK pin configuration required by ES7210
  - All new settings have default values, ensuring backward compatibility with other targets
- ES7210 initialization and I2S settings for M5Stack CoreS3 are based on the https://github.com/m5stack/M5Unified

#### M5Stack CoreS3 Target

- Update manifest.json audio input settings:
  - Change from PDM to I2S standard format
  - Set sample rate to 16000 Hz
  - Enable stereo (2 channels) with I2S_STD_SLOT_BOTH
  - Add MCK pin configuration (pin 0)

#### ES7210 Driver (setup-target.js)

- Add ES7210 class for microphone initialization
- Configure microphone gain, HPF, and power settings
- Enable AXP2101 ADC for voltage measurement

#### Audio Input Module (modules/io/audioin/esp32/audioin.c and modules/pins/audioin/esp32/audioin.c)

- Add MODDEF_AUDIOIN_I2S_MCK_PIN for master clock pin support (default: I2S_GPIO_UNUSED)
- Add MODDEF_AUDIOIN_I2S_FORMAT_I2S for I2S Philips format selection (default: 0)
- Add MODDEF_AUDIOIN_NUMCHANNELS for mono/stereo configuration (default: 1)
- Add MODDEF_AUDIOIN_I2S_SLOT for flexible slot selection (default: I2S_STD_SLOT_RIGHT)

### DIFF
--- a/build/devices/esp32/targets/m5stack_cores3/manifest.json
+++ b/build/devices/esp32/targets/m5stack_cores3/manifest.json
@@ -112,14 +112,18 @@
 				"dataout_pin": 13
 			}
 		},
-		"audioIn": {
-			"sampleRate": 11025,
+	"audioIn": {
+			"sampleRate": 16000,
 			"bitsPerSample": 16,
+			"numChannels": 2,
 			"i2s": {
+				"num": 1,
+				"slot": "I2S_STD_SLOT_BOTH",
+				"format_i2s": 1,
 				"bck_pin": 34,
 				"lr_pin": 33,
-				"datain": 14,
-				"pdm": 1
+				"mck_pin": 0,
+				"datain": 14
 			}
 		},
 		"camera": {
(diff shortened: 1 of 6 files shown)


phoddie:
The changes look good.

One suggestion. It is relatively expensive to convert from stereo to mono in JavaScript. An alternative would be for the audio input class to accept the `channels` argument on its constructor so that it could perform the conversion before passing the samples to the JavaScript callback.  We can merge this PR without that support. But if you'd like to add it to this PR, please let us know.

meganetaaan:
Thanks for the suggestion.

I found that simply using channels: 1 is sufficient in our case. AudioIn already implements stereo→mono conversion (it takes the L channel), so there’s no need for JS-side mixing. When I want stereo, channels: 2 already gives me stereo samples, which is good enough so far.
Based on this, I reverted the JS conversion in capture-sync and just set channels: 1.

phoddie:
@meganetaaan – thank you for making the change. Setting `channels` in the invocation of the constructor clearly expresses the requirement of the app for the samples to be a single channel.

mkellner:
This change has been merged and will be available in the next release.

PR #1554, opened 2026-01-05
Fix audioout ISR posting and queue-full handling (CoreS3)

meganetaaan:
* Moddable SDK: 6.1 + [cores3 microphone feature](https://github.com/Moddable-OpenSource/moddable/pull/1552)
* ESP-IDF: 5.5
* Device: `esp32/m5stack_cores3`

```
cd contributed/conversationalAI
mcconfig -m -i -p esp32/m5stack_cores3`
```

### Issue

ChatAudioIO frequently hangs while playing response audio (audio playback stops mid‑utterance).

### Root Cause
Based on a core dump provided by @stc1988 and subsequent investigation:

* playedBuffer runs in an ISR but called task-context APIs (postMessage / modMessagePostToMachine).
* Queue-full failures from ISR posting were ignored, leaving callbackPending stuck and preventing further delivery.

### Fix

* Use ISR‑safe APIs in the I2S callback (xTaskNotifyFromISR, modMessagePostToMachineFromISR).
* Return failure from modMessagePostToMachineFromISR when the queue is full.
* In audioout, detect this failure and roll back callbackPending/useCount so delivery can retry on the next IRQ.

### Expected Result

* Audio output no longer stalls mid‑utterance; ISR posting failures no longer dead‑lock delivery.

---

Notes / Verification

Confirmed fixed on M5Stack CoreS3. Not yet verified on other targets (e.g., moddable_six).
Concerned about potential impact on other builds/platforms; additional verification is appreciated.

### DIFF
--- a/modules/io/audioout/esp32/audioout-esp32.c
+++ b/modules/io/audioout/esp32/audioout-esp32.c
@@ -531,6 +531,7 @@ void xs_audioout_mark_(xsMachine* the, void* it, xsMarkRoot markRoot)
 static bool playedBuffer(i2s_chan_handle_t handle, i2s_event_data_t *event, void *user_ctx)
 {
 	AudioOut audioOut = user_ctx;
+	BaseType_t higherPriorityTaskWoken = pdFALSE;
 
 //@@ entire operation on audioOut->bytesWritable should be atomic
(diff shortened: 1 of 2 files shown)


phoddie:
Good catch! Those FreeRTOS ISR calls are something of a footgun. At first glance, your changes look quite reasonable. We'll do a more thorough review.

mkellner:

PR #1558, opened 2026-01-11
add typing embedded:io/bluetoothle/peripheral

stc1988:


typings/embedded_io/ble/server.d.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    interface GATTServerOptions {
+        mtu?: number;
+        services: GATTServerService[];
+        security?: {
+            authenticate?: boolean,
+            bond?: boolean,
+            immediate?: boolean,
+            ioCapabilities?: "none" | "display" | "numbers" | "display+numbers" | "display+confirm"
+        };
```
phoddie: This should be shared with client – they are identical

typings/embedded_io/ble/server.d.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        onRead?(this: GATTServerCharacteristic, connection: GATTServerConnection): Buffer;
+        onWrite?(this: GATTServerCharacteristic, value: ArrayBuffer, connection: GATTServerConnection): void;
+        onSubscribe?(this: GATTServerCharacteristic, connection: GATTServerConnection): void;
+        onUnsubscribe?(cthis: GATTServerCharacteristic, onnection: GATTServerConnection): void;
+    }
+
+    interface GATTServerDescriptor {
+        uuid: string;
+        value: Buffer;
```
phoddie: `value` is optional because onRead can be used to provide the value

typings/embedded_io/ble/server.d.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        onRead?(connection: GATTServerConnection): Buffer;
+        onWrite?(value: ArrayBuffer, connection: GATTServerConnection): void;
+    }
+
+    interface GATTServerAdvertisingRecords {
+        name?: string;
+        services?: string[];
+        manufacturerData?: { manufacturer: number, data: Buffer };
+        flags?: 1 | 2 | 4 | 8 | 16;
```
phoddie: I think the flags can be merged (e,g, 1+2) so maybe this needs to be a number instead of enumeration.

typings/embedded_io/ble/server.d.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    class GATTTServer {
+        constructor(options: GATTServerOptions);
+
+        onReady?(): void;
+        onConnect?(connection: GATTServerConnection): void;
+        onDisconnect?(connection: GATTServerConnection): void;
+        onPasskey?(connection: GATTServerConnection, action: "input" | "display" | "compareNumber" | "outOfBand", data?: number | ArrayBuffer): void;
+        onSecured(connection: GATTServerConnection, state: { authenticated: boolean, bonded: boolean, encrypted: boolean, keySize: number }): void;
```
phoddie: `state` could share `GATTSecurityState` with the client

typings/embedded_io/ble/server.d.ts
```diff
@@ -0,0 +1,96 @@
...
+        onWarning?(message: string): void;
+        close?(): void;
+        addService(service: GATTServerService): void;
+        deleteService(service: GATTServerService): void;
+        startAdvertising(scan: GATTServerAdvertisingRecords, response: GATTServerAdvertisingRecords): void;
```
phoddie: `response` is optional

typings/embedded_io/ble/peripheral.d.ts
```diff
@@ -0,0 +1,96 @@
+declare module "embedded:io/bluetoothle/peripheral" {
+    import type { Buffer } from "embedded:io/_common";
```
phoddie, line 2: Importing `Buffer` from _common is a little archaic at this point. `BufferLike` is preferred – it is ambiently defined in xs.d.ts. It is probably fine to leave this as-is for now. Longer term we should remove `Buffer` from `_common`, rename `BufferLike` to `ByteBuffer` to match the ECMA-419 term, and fully transition to `ByteBuffer`.

phoddie (commented):
Overall, this looks really good. Thank you for taking care of this. I can cross one item off my to-do list. ;)

I added some comments for your review and consideration. If you don't want to make them, we can merge this, and we'll address them (mostly by adding `ble/common.d.ts`).

typings/embedded_io/ble/peripheral.d.ts
stc1988: I leave this point as-is for now.

stc1988:
Thank you for the review. I made changes based on the review.

phoddie:
Wonderful, thank you! I think we can still get this merged for our (late...) January release.

stc1988:
OK. I started building the TypeScript application based on this type definition, but I found a few issues and fixed them. I’ve confirmed that the application can now be built.

mkellner:
This change will be available in the next release.

PR #1561, opened 2026-01-27
specify an API key in ChatAudioIO

stc1988:
Currently, the API KEY can only be specified from the `config`, but I have added an option to allow users to use their own specified values.　There is no impact on the existing application code.

Also, when I checked the behavior of this update, I found several issues, so I fixed them. Please see the commit for details.

### DIFF
--- a/modules/network/services/chatAudioIO/workers/humeAIEVI.js
+++ b/modules/network/services/chatAudioIO/workers/humeAIEVI.js
@@ -28,7 +28,6 @@ class HumeAIEVIModel extends ChatWebSocketWorker {
 	constructor(options) {
 		super(options);
 		this.host = "api.hume.ai";
-		this.path = `/v0/evi/chat?api_key=${config.humeAIKey}`;
 		this.headers = null;
 		this.audioPrefix = audioPrefix;
 		this.audioSuffix = audioSuffix;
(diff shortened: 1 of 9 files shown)


modules/network/services/chatAudioIO/workers/humeAIEVI.js
```diff
@@ -105,6 +101,8 @@ class HumeAIEVIModel extends ChatWebSocketWorker {
 				case 3: 
 					client.close();
 					this.path += `&config_id=${json.id}`;
+					this.headers.delete("content-length");
+					this.headers.delete("Content-Type");
```
phoddie, line 105: I'm curious about this change. I think I can see why you might need to clear `content-length` but isn't `Content-Type` always "application/json"?

stc1988: Indeed, the connection error was resolved simply by clearing the `content-length`. On the other hand, `Content-Type` is not considered required. 
I don’t think it will cause an error in the future by being considered unnecessary, but I cleared the non-required items

phoddie: Thank you for the explanation. We should be consistent in how we communicate with a esrvice.

With this change, we set the Content-Type for some requests and not others. If you are confident the Content-Type is optional, then we should remove it from the headers altogether - then there's no need to delete it. Still, my experience is that it is safer to explicitly set the Content-Type, but maybe that's no longer the case.

stc1988: I reviewed the code again, and with this change there is no modification to the headers used for Web API requests.
The update only removes headers from the WebSocket connection request.

stc1988: @phoddie 
If there are still any unclear points, please let me know.

mkellner:

PR #1566, opened 2026-02-19
Remove redundant _M_LN2 constant and alias

jplexer:
M_LN2 is now provided by the Pebble libc math headers, so the local define and its intermediate constant are no longer needed.


mkellner:
Merged and pushed.

PR #1572, opened 2026-03-15
Move ChatWebSocketWorker logic into model modules

stc1988:
Each AI model has many configuration options, but at present they cannot be changed through the configure method, and it is not practical to keep expanding the configure parameters each time.

So, in this PR, we export the models so that each AI model can be extended. This makes it possible to create models with customized configure behavior as shown below.

```js
import OpenAIRealTimeModel from "openAIRealtime";

class OpenAIRealTimeCustomModel extends OpenAIRealTimeModel {
	configure(message) {
		super.configure(message);
		this.session.audio.input = {
			format: { type: "audio/pcma" },
			noise_reduction: { type: "far_field" },
			transcription: { model: "gpt-4o-mini-transcribe", language: "ja" },
			turn_detection: {
				type: "semantic_vad",
				create_response: true
			}
		};
		this.session.audio.output.speed = 1.25
	}
}

new OpenAIRealTimeCustomModel({
	inputSampleRate: 8000
});


```


phoddie:
Thanks for the PR. It is a bit of a big change, so it may take us a while longer to review.

phoddie:
Thank you for your patience. We've reviewed this and tried it out. We agree with the approach and will merge this for our April release.

mkellner:

PR #1576, opened 2026-03-23
enhance memory safety in DNS parser by adding bounds checks

rootvector2:
Fix memory-safety issues in DNS parsing (bounds checks, pointer validation, integer underflow) to prevent out-of-bounds reads and malformed packet handling issues.

### DIFF
--- a/modules/network/dns/moddnsparser.c
+++ b/modules/network/dns/moddnsparser.c
@@ -50,20 +50,14 @@ static void *getQuestionByIndex(xsMachine *the, uint8_t index)
 
 	position += 12;
 	while (index--) {
-		while (true) {
-			if (position >= packetEnd)
-				return NULL;
+		while (position < packetEnd) {
 			uint8_t tag = *position++;
 			if (0 == tag)
 				break;
 			if (0xC0 == (tag & 0xC0)) {
-				if (position >= packetEnd)
-					return NULL;
 				position += 1;
 				break;
 			}
-			if (position + tag > packetEnd)
-				return NULL;
 			position += tag;
 		}
 		position += 4;
@@ -93,20 +87,14 @@ static void *getAnswerByIndex(xsMachine *the, uint8_t index)
 	while (index--) {
 		uint16_t rdlength;
 
-		while (true) {
-			if (position >= packetEnd)
-				return NULL;
+		while (position < packetEnd) {
 			uint8_t tag = *position++;
 			if (0 == tag)
 				break;
 			if (0xC0 == (tag & 0xC0)) {
-				if (position >= packetEnd)
-					return NULL;
 				position += 1;
 				break;
 			}
-			if (position + tag > packetEnd)
-				return NULL;
 			position += tag;
 		}
 		position += 10;
@@ -127,13 +115,15 @@ static void *getAnswerByIndex(xsMachine *the, uint8_t index)
 static int parseQname(xsMachine *the, int offset)
 {
 	uint8_t *packetEnd;
-	uint8_t *packetStart = getPacket(the, &packetEnd);
-	xsUnsignedValue packetLength = (xsUnsignedValue)(packetEnd - packetStart);
+	uint8_t *packetStart;
+	xsUnsignedValue packetLength;
 	uint8_t *position;
 	int qnameEndPosition = 0;
 	int hops = 0;
 
-	xsVar(1) = xsNewArray(0);
+	xsVar(1) = xsNewArray(0);	// do this before getPacket: xsNewArray can trigger GC
+	packetStart = getPacket(the, &packetEnd);
+	packetLength = (xsUnsignedValue)(packetEnd - packetStart);
 	position = offset + packetStart;
 	while (true) {
 		char tmp[64];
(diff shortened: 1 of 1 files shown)


modules/network/dns/moddnsparser.c
```diff
@@ -107,37 +122,57 @@ static void *getAnswerByIndex(xsMachine *the, uint8_t index)
...
+	uint8_t *packetEnd;
+	uint8_t *packetStart = getPacket(the, &packetEnd);
+	xsUnsignedValue packetLength = (xsUnsignedValue)(packetEnd - packetStart);
 	uint8_t *position;
 	int qnameEndPosition = 0;
+	int hops = 0;
 
 	xsVar(1) = xsNewArray(0);
-	position = offset + (uint8_t *)getPacket(the, NULL);;
+	position = offset + packetStart;
```
phoddie, line 127: This change is unsafe because `xsNewArray` can cause a garbage collection which will move memory, potentially invalidating `packetStart`. The solution is probably to move `xsNewArray` to the start of the function.

modules/network/dns/moddnsparser.c
```diff
@@ -107,37 +122,57 @@ static void *getAnswerByIndex(xsMachine *the, uint8_t index)
...
 		if (tag > 63)
 			xsUnknownError("bad name");
+		if (position + tag > packetEnd)
+			xsUnknownError("bad name");
 		c_memcpy(tmp, position, tag);
 		xsmcSetStringBuffer(xsVar(2), tmp, tag);
 		xsCall1(xsVar(1), xsID_push, xsVar(2));
 		offset += tag;
-		position = offset + (uint8_t *)getPacket(the, NULL);
+		position = offset + packetStart;
```
phoddie, line 168: This change is unsafe because both `xsmcSetStringBuffer` and `xsCall` may cause memory to move invalidating `packetStart`.

modules/network/dns/moddnsparser.c
```diff
@@ -51,13 +51,19 @@ static void *getQuestionByIndex(xsMachine *the, uint8_t index)
 	position += 12;
 	while (index--) {
 		while (true) {
```
phoddie: I **think** the changes to this loop could be consolidated by changing the `while (true)` to `while (position < packetEnd)`. Then the check outside the loop (line 64) will always pick up the malformed packet.

phoddie (changes requested):
Thanks for the PR to harden the DNS parsing. Very much appreciated. I left a few notes for changes. In particular, it is important to keep in mind that the garbage collector in the XS engine compacts memory, which can invalidate pointers held in C. 

Before we can merge this PR, we need you to complete our [Contributor License Agreement](https://github.com/Moddable-OpenSource/moddable/tree/public/licenses) - choose either the Individual or corporate, as appropriate.

rootvector2:
I’ve completed the CLA (individual). Please let me know where to send the signed copy.

phoddie:
> I’ve completed the CLA (individual). Please let me know where to send the signed copy.

Please send a scan (preferably PDF) by email to info@moddable.com. Thank you.

rootvector2:
> > I’ve completed the CLA (individual). Please let me know where to send the signed copy.
> 
> Please send a scan (preferably PDF) by email to [info@moddable.com](mailto:info@moddable.com). Thank you.

ok Sent

phoddie:
Thanks – we received the CLA.

There is an inconsistency. Most of the changes throw when invalid data is detected. I think that's the right behavior. The one exception is the `txt` parsing – which exits early without reporting an error. That means that the script is unaware that it is working with incomplete, possible corrupt, information. Is there a reason for the different behaviors? If not, my feeling is that the `txt` parsing should throw when it detects incorrect data.

rootvector2:
I agree that silently exiting in TXT parsing can lead to incomplete or misleading data being returned to the script. It should behave consistently with the rest of the parser and throw on invalid input.

I’ll update the TXT parsing to throw an error instead of breaking.

phoddie:
Great, thank you. We should have everything needed to merge this as part of our upcoming April release.

rootvector2:
Thanks, really appreciate the review and guidance!

If there’s anything else I can help with or take up, feel free to let me know.

phoddie:
Happy to have the help! I am curious... the parsing errors are real but a bit obscure. How did you stumble over them in the first place?

rootvector2:
Honestly, I was just going through the DNS parser with a focus on how it handles malformed inputs and memory safety. I noticed a few TODO comments and some places where bounds checks were missing, especially around label parsing and compression pointers.

That got me curious, so I followed those paths and tried to think through how malformed packets might behave, which is how I ended up spotting those cases.

phoddie:
Very cool! If you are looking for a good challenge, staying within the DNS module the serializer (`dnsserializer.js`) takes a shortcut storing names that results in bigger packets and, I believe, may be strictly non-conforming. The names passed to `add` are supposed to appear in the serialized packet just once, with occurrences beyond the first pointing back to the original. The requires some additional bookkeeping during serialization. If you look at the way the [DNS-SD code](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/io/dnssd/dnssd.js) uses the serialized to construct packets, you can see how the same name will appear in several different parts of its packets.

rootvector2:
any update?

phoddie:
Welcome back. It looks like the [moddnsparser.c](https://github.com/Moddable-OpenSource/moddable/blob/public/modules/network/dns/moddnsparser.c) changes landed. The most recent commit :
> [rootvector2](https://github.com/Moddable-OpenSource/moddable/commits?author=rootvector2) and [mkellner](https://github.com/Moddable-OpenSource/moddable/commits?author=mkellner) [bounds-check dns response parsing against packet end](https://github.com/Moddable-OpenSource/moddable/commit/8a3395e9c1d87f76f57f66c561b6a625117e08b1)

rootvector2:
just saw the commit, thanks for landing this — glad the `moddnsparser.c` hardening made it in.

phoddie:
Thanks again for the help. Safe handling of DNS input is a welcome improvement since it is a pervasive attack vector. Looking forward to what you uncover next.

PR #1577, opened 2026-03-24
fix(pico): resolve build errors for GCC 14+, prebuilt picotool

HipsterBrown:
While implementing https://github.com/HipsterBrown/xs-dev/pull/248, I ran into issues trying to build the "hello world" example program. After doing some debugging, I ended up with this working solution. 

The missing c_abort macro will make sure anyone using > version 14 of the ARM GCC toolchain can build programs targeting the pico. The `UF2CONV` enables developers to use prebuilt versions of `picotool` that can be placed in a different directory from the default CMake build path.


mkellner:

PR #1582, opened 2026-03-30
DNS: compress repeated names during serialization

rootvector2:
This PR updates DNS serialization so repeated names are encoded once, with later occurrences referencing the original through DNS compression pointers.

The previous serializer path could duplicate repeated names, which increases packet size and may produce non-conforming output in DNS-SD-style packets where the same names appear across multiple records.

## What Changed

- Added name offset bookkeeping during serialization.
- Reused previously written name suffixes via DNS compression pointers.
- Applied compression behavior to names in:
  - owner names
  - `PTR` RDATA names
  - `SRV` target names
  - `NSEC` next-domain names
- Kept the existing serializer API and record conversion behavior intact.

### DIFF
--- a/modules/network/dns/dnsserializer.js
+++ b/modules/network/dns/dnsserializer.js
@@ -36,13 +36,22 @@ class Serializer {
 		if (dictionary.id)
 			this.id = dictionary.id;
 	}
-	add(section, name, type, clss, ttl, data) {
-		const record = [];
+	splitName(name) {
+		if (!name)
+			return [];
 
+		const parts = name.split(".");
+		if ((parts.length > 1) && ("" === parts[parts.length - 1]))
+			parts.pop();
+		return parts;
+	}
+	add(section, name, type, clss, ttl, data) {
 		if (!data)
 			;
 		else if (data instanceof ArrayBuffer)
 			data = new Uint8Array(data);
+		else if (undefined !== data.byteLength)
+			data = new Uint8Array(data.buffer, data.byteOffset, data.byteLength);
 		else {
 			let d;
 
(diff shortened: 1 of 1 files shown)


phoddie:
Thanks for working on this! The changes appear to be working correctly. I verified the binary packets using $MODDABLE/examples/io/dnssd.

In walking through the code, I made a few changes mostly to tighten up the code. The two main points are:

- No need to `& 255` when assigning to a `Uint8Array` entry since the language already does that for you
- It is noisy to pass `state` around to so many functions. I moved it to `this`.

The full set of changes is in a Gist. You can see the [diff of all the changes here](https://gist.github.com/phoddie/20a6a4b6b55a7dc2bfc8faabfcfad5b8/revisions). Please let me know what you think.

rootvector2:
Thanks for the review and for sharing the gist. I applied the updates on the branch and checks are still passing. Looks good from my side. Let me know if anything else needs to be done.

rootvector2:
@phoddie ?

phoddie:
Hey, thanks for merging the suggestions. Since you opened up this code to make some functional improvements, it was a good opportunity to clear up a few things. We'll get this merged for our April SDK release.

There are lots of other things that could benefit from some attention. Let me know a bit about your areas of interest (beyond DNS!) and environment (Mac, Linux, or Windows host) and embedded hardware (ESP32, nRF52, Zephyr, etc) so I can make some suggestions. Thanks!

rootvector2:
I’m on macOS using the simulator and host setup
I don’t have hardware right now so mainly working on things I can test on host

Beyond DNS I’m interested in networking side work like making parsers and serializers more robust improving memory safety in C code and handling edge cases properly

But I’m also completely open to other areas if you think something else would be more useful I’m happy to pick it up and learn as I go

phoddie:
Hey, apologies for being slow to reply. Things have been busy between last week's [Zephyr Tech Talk](https://www.youtube.com/watch?v=SCl_EAwHv2Y) and working on our upcoming April Moddable SDK release (which will incorporate your changes).

For the most part, all the parsing and serializing of data for network protocols is done JavaScript rather than C in the Moddable SDK. We do that for many reasons:

- JavaScript is memory safe (by definition!), so an exception is the worst that can happen with unexpected data.
- JavaScript is portable.
- Parsing and serializing text is much easier in JavaScript than C. (Binary data in JavaScript is sometimes a headache though).
- JavaScript is fast enough for these operations.

The DNS parser C code you hardened is an exception. It is a fair question whether we should simply migrate it to JavaScript for safety.

If you look at our core ECMA-419 IO support (`$MODDABLE/modules/io/**/*.c`), you'll see that there's not a C parser or serializer in sight. As the bridge between JavaScript and hardware, we depend on that layer to be memory safe. If there are issues there, they are high priority to resolve. The code is very platform dependent – macOS TCP looks very little like ESP32 TCP, for example - though the JavaScript APIs are the same. You might take a look there – you might spot something we overlooked or deferred for later.

FWIW – we do continuous fuzz testing of our XS JavaScript engine using both Fuzzilli and oss-fuzz ([some background](https://www.moddable.com/hardening-xs)). That's helped to make XS itself reliably memory safe. But we don't do the same for the many modules that make up the runtime. We can't fuzz on ESP32, but it should be possible to do that for Commodetto / Poco and Piu, for example, as they are 99% portable C code. But that's a big complicated project.

An interesting place to start might simply be to look at the Commodetto / Poco and Piu APIs for vulnerabilities. That could be done by code inspection, a test harness, or a combination of both. The most interesting issues are those that lead to memory safety failures. Here's [one place](https://github.com/Moddable-OpenSource/moddable/blob/fefb4aabee9a9cfe159efad220cd1909c11f6d5d/modules/commodetto/cfeBMF.c#L77) that is certainly capable of an out-of-bounds read with malformed data. On macOS, it is possible to enable ASan (address sanitizer) in the build to aggressively check for out-of-bounds reads and writes (it isn't a build option, so you'd need to experiment a bit to get it enabled).

Let me know if any of that sounds interesting or you have questions. This is an interesting and important area, but not exactly simple.

phoddie:
The reported issue is resolved, so closing this out. Any thoughts on the notes above from about two weeks ago?

rootvector2:
Hi @phoddie — apologies for going quiet here. Some things came up on my
end and I had to step away for a bit, but I'm back and planning to start
working.

PR #1584, opened 2026-04-01
simulator: fix screen_begin for ECMA-419 display calling convention

### LINKED ISSUE #1579: Commodetto/Poco use is broken for the simulator
**Build environment:** macOS
**Moddable SDK version:** >= v7
**Target device:** desktop simulator

**Description**
Attempting to run any program using Poco ("commodetto/Poco" module) in the simulator results in an error:
```
Break: SyntaxError: C: xsArg(1): invalid index (in @screen_begin)!
```
It will run fine on a device, i.e. Moddable Six, but not the simulator version. 

**Steps to Reproduce**
1. `cd $MODDABLE/example/commodetto/text`
2. `mcconfig -d -m -p sim/moddable_six`

**Expected behavior**
Simulator starts with the demo text appearing on the screen

**Images**
Image

**Other information**
I've debugged this down to the ECMA-419 display support being added in v7 of the SDK. The `manifest_commodetto.json` defaults to defining `ECMA419.display: 1` which should use that branch of logic when rendering graphics with Poco. When I set that flag to `0`, then the example runs successfully. This leads me to believe the `screen` module for the simulators wasn't updated to match the expected logic for the ECMA-419 Display implementation, starting with somewhere in the `screen_begin` method: https://github.com/Moddable-OpenSource/moddable/blob/public/build/simulators/modules/screen.c#L625

HipsterBrown:
Fixes #1579.

`manifest_commodetto.json` includes `$(MODULES)/io/display/manifest.json`, which sets `ECMA419.display: 1` and compiles Poco with `MODDEF_ECMA419_DISPLAY=1`. Under this path, `commodettoPocoCore.c` calls `pixelsOut.begin()` with a single object argument {x, y, width, height} rather than four separate integers.

The simulator's screen_begin was never updated to handle this calling convention. It unconditionally read `xsArg(0)`, `xsArg(1`), and `xsArg(2`) as integers. When called with one object argument, `xsArg(1)` is out of bounds, producing:

```
Break: SyntaxError: C: xsArg(1): invalid index (in @screen_begin)!
```

Real hardware (e.g. Moddable Six) is unaffected because its display driver uses native C function pointer dispatch (displayHooks / pixelsOutDispatch), which bypasses the JavaScript begin() call entirely.

The fix checks the argument count in `screen_begin`: if one argument is passed, it reads x, y, and width from the object's properties; otherwise it falls through to the original integer-argument path, preserving backward compatibility.

### DIFF
--- a/build/simulators/modules/screen.c
+++ b/build/simulators/modules/screen.c
@@ -625,9 +625,17 @@ void screen_animateColors(xsMachine* the)
 void screen_begin(xsMachine* the)
 {
 	txScreen* screen = xsGetHostData(xsThis);
-	xsIntegerValue x = xsToInteger(xsArg(0));
-	xsIntegerValue y = xsToInteger(xsArg(1));
-	xsIntegerValue width = xsToInteger(xsArg(2));
+	xsIntegerValue x, y, width;
+	if (1 == xsToInteger(xsArgc)) {
+		x = xsToInteger(xsGet(xsArg(0), xsID_x));
+		y = xsToInteger(xsGet(xsArg(0), xsID_y));
+		width = xsToInteger(xsGet(xsArg(0), xsID_width));
+	}
+	else {
+		x = xsToInteger(xsArg(0));
+		y = xsToInteger(xsArg(1));
+		width = xsToInteger(xsArg(2));
+	}
 	screen->rowAddress = screen->buffer + (y * screen->width * screenBytesPerPixel) + (x * screenBytesPerPixel);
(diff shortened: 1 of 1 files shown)


phoddie:
Thank you for tracking this down! The change makes sense. I have a vague memory of making that change but, even if that memory is true, it it clearly never made it to the repo.

We'll get this merged for the April release.

HipsterBrown:
Closing this as the fix has been released in v8.0.0

PR #1585, opened 2026-04-02
Add typing for mod/config

stc1988:


phoddie:
Thanks for adding this. It makes sense to have `mod/config` and have it match `mc/config`. We'll get this merged.

PR #1586, opened 2026-04-04
feat: enable Linux simulator AudioIn and AudioOut linking

meganetaaan:
## Summary

  Enable `embedded:io/audio/in` and `embedded:io/audio/out` on the Linux simulator by linking ALSA only when those modules are used.

  ## Changes

  - add optional Linux audio build variables to `tools/mcconfig/make.lin.mk`
  - request `-lasound` from the Linux `AudioIn` manifest
  - request `-lasound` from the Linux `AudioOut` manifest
  - add `linux-sim-audio.md` with the design note for Linux simulator audio support

  ## Why

  Linux already has ALSA-based implementations for `AudioIn` and `AudioOut`, but simulator apps were not linking against ALSA. As a result, apps using these modules could fail at link time
  or load time with unresolved audio symbols.

  This change wires the existing Linux audio implementations into simulator builds without changing the JS API and without adding ALSA to unrelated simulator apps.

  ## Verification

  Verified on top of the latest `upstream/public`.

  - built `examples/io/audioout/play-sync` for `-p sim/moddable_six`
  - built `examples/io/audioin/capture-sync` for `-p sim/moddable_six`
  - confirmed generated `mc.so` links `libasound.so.2`
  - launched `play-sync` in `mcsim`
  - launched `capture-sync` in `mcsim` and confirmed `recording -> playing -> done`

### DIFF
--- a/modules/io/audioin/manifest.json
+++ b/modules/io/audioin/manifest.json
@@ -10,6 +10,9 @@
 			}
 		},
 		"lin": {
+			"build": {
+				"LINUX_AUDIO_LIBS": "-lasound"
+			},
 			"modules": {
 				"*": "$(MODDABLE)/modules/io/common/builtinCommon",
 				"embedded:io/audio/in": "./lin/audioin"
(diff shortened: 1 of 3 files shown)


mkellner:
This will be in the next release. I generalized it to `C_FLAGS` and `LIBRARIES` so other modules can add libraries.

PR #1592, opened 2026-04-12
fix(io/spi/esp): conditionally call modSPITxSwap16 for relevant displays

### LINKED ISSUE #1591: embedded:io/spi fails to link on ESP8266 with non-RGB565LE display formats
**Build environment:** macOS
**Moddable SDK version:** 7.0.0
**Target device:** Moddable Three

**Description**
Including $(MODDABLE)/modules/io/manifest.json in a project targeting any ESP8266 board whose display format is not RGB565LE (e.g. Moddable Three, which uses Gray256) produces a linker error for the undefined symbol `modSPITxSwap16`.

**Steps to Reproduce**
1. Create a project whose manifest.json includes $(MODDABLE)/modules/io/manifest.json
2. Build for esp8266/moddable_three: `mcconfig -d -m -p esp8266/moddable_three`
3. See linker error:
```
  modules/io/spi/esp/spi.c: undefined reference to `modSPITxSwap16'
  collect2: error: ld returned 1 exit status
```

**Expected behavior**
The project links and flashes successfully.

**Images**
N/A

**Other information**
`modules/io/spi/esp/spi.c` calls `modSPITxSwap16` unconditionally when spi->transform is set. However, in `modules/pins/spi/esp/modSPI.c` the function is only compiled when `kCommodettoBitmapFormat == kCommodettoBitmapRGB565LE`. On Moddable Three the format is `Gray256`, so the symbol is never emitted and the link fails.

Suggested fix — guard the call in spi.c with the same #if already used in modSPI.c, falling through to the plain modSPITx on non-RGB565LE targets:

```c
// modules/io/spi/esp/spi.c  xs_spi_write
#if kCommodettoBitmapFormat == kCommodettoBitmapRGB565LE
      if (spi->transform)
              modSPITxSwap16(&spi->config, (uint8_t *)data, (uint16_t)count);
      else
#endif
              modSPITx(&spi->config, (uint8_t *)data, (uint16_t)count);
```

HipsterBrown:
Fixes #1591

### DIFF
--- a/modules/io/spi/esp/spi.c
+++ b/modules/io/spi/esp/spi.c
@@ -253,9 +253,11 @@ void xs_spi_write(xsMachine *the)
 	if (count > 65535)
 		xsRangeError("unsupported byteLength");
 
+#if kCommodettoBitmapFormat == kCommodettoBitmapRGB565LE
 	if (spi->transform)
 		modSPITxSwap16(&spi->config, (uint8_t *)data, (uint16_t)count);
 	else
+#endif
 		modSPITx(&spi->config, (uint8_t *)data, (uint16_t)count);
 }
 


phoddie:
Thanks for tracking this down and providing a fix. Coincidentally, I've been working on a tool to test our examples more thoroughly. Had I run it again Moddable Three, it would have caught this. 

I wonder if real mistake is in `modSPI.c`. That could always provide `modSPITxSwap16` (and more obscure friends like `modSPITxRGB332To16BE`). When they are unused, the linker will remove them.

HipsterBrown:
> Coincidentally, I've been working on a tool to test our examples more thoroughly. Had I run it again Moddable Three, it would have caught this.

As more examples and APIs get added, that sounds like a very useful tool. It makes me think of the [ecosystem CI for Vite](https://github.com/vitejs/vite-ecosystem-ci) but scoped to this repo. I'm interested to hear more about what kind of canary devices / device farm you might be setting up to support this as well. 

> I wonder if real mistake is in modSPI.c. That could always provide modSPITxSwap16 (and more obscure friends like modSPITxRGB332To16BE). When they are unused, the linker will remove them.

That would certainly solve the issue too. My initial local solution was to just stub `modSPITxSwap16` in a C file to include in my project build. If there's no build or runtime cost to defining those functions, it could help simplify `modSPI.c` along with fixing this bug. 

Btw it's been a lot of fun working with the Moddable Three again. I'm building a little climate monitor for my 3D printer filament storage to integrate with Home Assistant. It would be neat to see a refreshed version based on an RP2040 or RP2350 (with or without WiFI). 😄

phoddie:
I made the change to modSPI.c described above. I tested a variety of devices and examples. It seems solid. Fingers crossed.

> I'm interested to hear more about what kind of canary devices / device farm you might be setting up to support this as well

Me too. ;) It is currently ad-hoc – whatever devices happen to be on-hand. That's already proved very effective at identifying issues.

One annoyance is that sometimes the serial connection and/or the device just get in a bad state. It is rare but it stalls testing. That may be out of scope for this tool though. At best, it can try to detect that.

I think the tool needs to be modified to generate a separate report.json from each run, rather than overwriting the last one. That would make it possible to do lots of runs and then generate a summary of the results. More automation!

> Btw it's been a lot of fun working with the Moddable Three again. I'm building a little climate monitor for my 3D printer filament storage to integrate with Home Assistant. It would be neat to see a refreshed version based on an RP2040 or RP2350 (with or without WiFI). 😄

Great to hear. Moddable Three is wonderful – the device is just the size of the screen. ePaper displays are better than LCDs for many situations. The ESP8266 is often all the horsepower you need (and build / install / run cycles are fast). Using an RP would bring a lot more power. The biggest headache is the display – each time we order another batch of screens, there's an excellent chance the driver needs to be tweaked.

PR #1593, opened 2026-04-16
fix how to running test-examples

stc1988:
I fixed how to running test-examples and add tools/test-examples/report.json to .gitignore.

I hava a question about below Note:

```
Note: test-examples is implemented using Node. Before running it for the first time, you must execute "npm install" in the $MODDABLE/tools/test-examples directory.
```

There is no `package.json`, and if test-examples uses xsdb, I think `npm install` must be run in the `$MODDABLE/tools/xsbug-log` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ build/devices/qca4020/xsProj/build/gcc/app.*
 build/devices/qca4020/xsProj/build/gcc/dbg.*
 build/devices/qca4020/xsProj/build/gcc/dictLog
 build/devices/qca4020/xsProj/build/gcc/output/
+tools/test-examples/report.json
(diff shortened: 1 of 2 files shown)


phoddie:
Thank you for the corrections.

Even though there are currently no dependencies, it seems possible there will be in the future. I added simple package.json so that the tool is clearly marked as a Node project.

It is true that xsdb also needs an "npm install" to use "test-examples". We should either:

- Automatically do "npm install" for xsdb when doing it for "test-examples"
- Update the "test-examples" document to note the requirement to also "npm install" for xsdb

phoddie:
This will be merged in the next Moddable SDK release.

PR #1598, opened 2026-04-20
pebble/touch: use applib touch_service API

gmarull:
Replace the services/common/event_service and drivers/touch/touch_sensor usage with the applib touch_service API. The applib wrapper delivers TouchEvent directly to the handler and manages the touch sensor lifecycle via subscriber-count tracking, so the per-instance EventServiceInfo, PebbleEvent unwrapping, and explicit touch_sensor_set_enabled/touch_reset calls are no longer needed.

Adapts to pebbleos commit 9e8982d5d which simplified the touch event model (TouchEvent now carries type + int16 x/y directly, and touch_dispatch_touch_events/TouchIdx were removed).

### DIFF
--- a/build/devices/pebble/modules/touch/pebble-touch.c
+++ b/build/devices/pebble/modules/touch/pebble-touch.c
@@ -24,12 +24,7 @@
 
 #include "builtinCommon.h"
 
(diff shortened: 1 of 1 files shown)


mkellner:

PR #1600, opened 2026-04-21
M5atom Matrix and S3 migrate MPU6886 from pins/i2c to Ecma-419

stc1988:
Mixing pins/i2c with ECMA-419 I²C on the devices causes a conflict error.

Therefore, as a first step, this PR migrates devices that require minimal changes to ECMA-419 I²C. Please let me know if you have any concerns.


phoddie:
```js
[sample.gyroscope.x, sample.gyroscope.y] = [sample.gyroscope.y * -1, sample.gyroscope.x];
```

That's a subtle way to get a hidden temporary variable. ;)

stc1988:
ChatGPT is much more of a JavaScript expert than I am.

phoddie:
For those of your following along at home, here's [a backgrounder](https://medium.com/@rolandogbue/understanding-destructuring-vs-swapping-in-javascript-a79f23ea2536).

stc1988:
In this PR, the main focus is on resolving the I2C conflict, but if we are to fully migrate to ECMA-419, I think it would be necessary to create common modules for things like buttons as well. What do you think?

phoddie:
> ...if we are to fully migrate to ECMA-419...

We are. Still considering the  time line, but I expect it to be this calendar year. (Don't panic... it will not be a surprise!)

> ...I think it would be necessary to create common modules for things like buttons as well. What do you think?

Absolutely.

mkellner:

PR #1605, opened 2026-05-07
Add rotation support to CO5300 display driver, and BMI270 IMU support for M5AtomS3R targets

kitazaki:
## Summary

Add rotation support to the CO5300 display driver used by `esp32/waveshare_amoled_206`.

The driver now supports `rotation` values of `0`, `90`, `180`, and `270`, consistent with other Moddable display drivers.

## Changes

- Added a `rotation` getter/setter to `modules/drivers/co5300/co5300.js`.
- Added native rotation state handling in `modCo5300.c`.
- Updated `width` and `height` getters for 90/270 degree rotation.
- Added validation for unsupported rotation values.
- Added software pixel rearrangement for 90, 180, and 270 degree rotation.
- Preserved the existing fast path for 0 degree rotation.

## Notes

The CO5300 does not appear to provide a MADCTL-style row/column exchange mode suitable for 90/270 degree rotation. Non-zero rotations are therefore handled in software before pixel data is sent to the display.

Rotated modes may be slower than the default 0 degree mode, but existing behavior is unchanged when `rotation` is not set.

## Testing

Tested on `esp32/waveshare_amoled_206`, which uses the CO5300 display driver.

Verified the following rotation values with `examples/piu/cards`:

- `0`
- `90`
- `180`
- `270`


kitazaki:
## Summary

This PR adds BMI270 IMU support to the M5AtomS3R target family.

Moddable SDK 8.1.0 added support for the BMI270 sensor device.
Since the M5AtomS3R family uses the same BMI270 IMU, this PR updates the corresponding target definitions.

Updated targets:

- `esp32/m5atom_s3r`
- `esp32/m5atom_s3r_cam`
- `esp32/m5atom_s3r_m12`

## Changes

- Added the BMI270 sensor driver manifest to the M5AtomS3R target manifests.
- Added `device.sensor.IMU` definitions for M5AtomS3R targets.
- Added axis correction for `m5atom_s3r` so the IMU orientation matches `m5atom_s3`.
- Scaled BMI270 accelerometer values from `m/s^2` to `g` for compatibility with existing AtomS3 IMU(MPU6886) examples.
- Updated `examples/drivers/atoms3-imu` to handle samples that may not include accelerometer or gyroscope data immediately.
- Added `examples/drivers/atoms3r-imu-console` for console-based IMU verification on displayless M5AtomS3R variants.

## Testing

- `examples/drivers/atoms3-imu` with `esp32/m5atom_s3r`
- `examples/drivers/atoms3r-imu-console` with `esp32/m5atom_s3r`
- `examples/drivers/atoms3r-imu-console` with `esp32/m5atom_s3r_cam`
- `examples/drivers/atoms3r-imu-console` with `esp32/m5atom_s3r_m12`




## Notes

M5Stack Core2 v1.3 also uses the BMI270 IMU. However, the Core2 hardware configuration changed between v1.1 and v1.3: the IMU changed from MPU6886 to BMI270
a separate PR for Core2 v1.3 after verifying the detection logic for these hardware differences.

stc1988:
@kitazaki 
Thank you for checking on multiple devices.

> Added examples/drivers/atoms3r-imu-console for console-based IMU verification on displayless M5AtomS3R variants.

If it’s displayless and doesn’t depend on any specific device, you might try creating examples/drivers/sensors/bmi270. That location makes it easier to verify the driver from other devices as well.

https://github.com/Moddable-OpenSource/moddable/tree/public/examples/drivers/sensors/mpu9250 can serve as a reference. For this device, polling alone might be sufficient.

phoddie:
> The CO5300 does not appear to provide a MADCTL-style row/column exchange mode suitable for 90/270 degree rotation. Non-zero rotations are therefore handled in software before pixel data is sent to the display.

Are you sure about this? The controller appears to be rooted in MIPI. It has MADCTL (0x36), which seems like it might work. Please see section 7.5.29 of the [data sheet](https://admin.osptek.com/uploads/CO_5300_Datasheet_V0_00_20230328_07edb82936.pdf). If it can work, it would be simpler and faster. Perhaps it is the absence of `MV` support that is limiting this? At least 180 rotations should be able to bypass the transformation buffer.

If the MADCTL really cannot be used for 90 and 270, we should optimize the transformation implementation because it could be faster. I can share some suggestions.

kitazaki:
@phoddie 

Thank you for pointing this out. I rechecked section 7.5.29 of the CO5300 datasheet.

MADCTL (0x36) is present, but the parameter only defines MY, MX, and RGB.
The MV bit does not appear to be defined for this controller, so I don’t think MADCTL alone can support 90 and 270 degree rotations because those require swapping row/column addressing.

However, I agree that 180 degree rotation should be possible using MY + MX.
I updated the implementation to use MADCTL for 180 degree rotation and keep the transformation path only for 90/270 degree rotations.

kitazaki:
@stc1988 

Thank you for your comment.
I moved the sample to `examples/drivers/sensors/bmi270/poll` and updated to use the BMI270 driver directly.
It uses `device.I2C.internal` when available and falls back to `device.I2C.default`, so it can be used from devices where the BMI270 is on an internal I2C bus as well as from external wiring setups.

phoddie:
@kitazaki – The change for 180 looks good. I used Moddable Two, which has a MIPI display controller, to experiment with `MY` and `MX` to remind myself how those work. You are correct that without the `MV` we cannot do 90 and 270 purely at the display controller level. Too bad.

But... I think the 90/270 conversion can be improved. Each transaction is relatively expensive – starting/stopping a transaction takes some time. Consequently, fewer transactions have less overhead.

- I'm not certain, but in `co5300SendRotated()` it seems possible that `co5300SetWindow()` can be moved outside the loop, so it only happens once. It is fine to do multiple writes.
- The rotation buffer could be bigger. It is only two scan lines. We try to do more like 16 at a time. (Also, the driver could wait to allocate the rotation buffer until needed, so apps running at 0 or 180 don't lose that memory.)
- The loop assumes that it can write either one or two lines (`rows = (lines > 1) ? 2 : 1;`). But... the number of lines depends on the width of the lines as passed to `co5300Begin()`.  Because Piu uses partial updates aggressively, often the width is much less than the full width. That means that the allocated rotation buffer can hold many more smaller lines than two full lines.
- It would be trickier, but you could have an asynchronous version of `co5300SendColorSync()`. That would allow some parallelism. I'm not sure how big a win it is compared to the complexity, but the 0 case is async and that helps nicely on Moddable Six (the driver that this code is modeled on).

kitazaki:
@phoddie 

Thank you for the detailed feedback.

I agree with your points. In particular, moving co5300SetWindow() outside the loop and increasing the rotation buffer size should reduce transaction overhead and improve performance for 90/270 rotation.

I will create a small benchmark sample program, test it on the actual device, and share the results. Based on that, I will try to update the implementation with the following improvements:

- Move co5300SetWindow() outside the loop if possible
- Increase the rotation buffer size
- Adjust the number of rotated lines dynamically based on the update width
- Consider allocating the rotation buffer only when 90/270 rotation is used

The asynchronous version of co5300SendColorSync() also sounds interesting, but I think I will first measure the simpler improvements above before adding that complexity.

Please give me a little time to test this.

kitazaki:
@phoddie 

## Summary

This improves the CO5300 software rotation path used for 90 and 270 degree rotations.

The CO5300 controller can handle 180 degree rotation with MADCTL MY/MX, but 90 and 270 degree rotations still require software rotation because MV does not appear to be supported. This change keeps that behavior and optimizes the software rotation implementation.

Changes:

- Increase the software rotation buffer from 2 scan lines to 16 scan lines.
- Allocate the rotation buffer lazily, only when 90 or 270 degree rotation is selected.
- Process as many source rows as fit in the rotation buffer based on the current update width, instead of always processing only one or two rows.
- Keep `co5300SetWindow()` inside the rotated chunk loop. Each chunk is emitted as a rotated vertical strip, so moving the window setup outside the loop would require a different packed pixel order.
- Add `examples/drivers/co5300-rotation-benchmark` to measure CO5300 rotation performance on device.

Hardware benchmark results on the CO5300:

| Rotation | Update area | Before | After | Improvement |
| --- | --- | ---: | ---: | ---: |
| 0 | full width | 11 ms/frame | 11 ms/frame | no change |
| 180 | full width | 13 ms/frame | 13 ms/frame | no change |
| 90 | full width | 40 ms/frame | 20 ms/frame | 2.0x faster |
| 270 | full width | 40 ms/frame | 20 ms/frame | 2.0x faster |
| 90 | narrow partial | 40 ms/frame | 5-6 ms/frame | about 6.7x-8.0x faster |
| 270 | narrow partial | 33 ms/frame | 5 ms/frame | about 6.6x faster |

Attached short before/after videos for visual comparison:
- `before_small.mp4`: before this optimization

https://github.com/user-attachments/assets/7cbbc3fc-ef7d-4258-9027-faf45cc5351b

- `after_small.mp4`: after this optimization

https://github.com/user-attachments/assets/9c877067-6e1d-4885-87e0-ebdfcf0927ad

kitazaki:
## Summary

This updates the `esp32/m5stack_core2` device definition to support both M5Stack Core2 hardware revisions.
Core2 v1.0/v1.1 uses MPU6886, while Core2 v1.3 uses BMI270.
This change adds runtime IMU detection so the device definition can select the appropriate driver based on the detected chip ID.

Changes:
- Add runtime detection for MPU6886 and BMI270 on I2C address `0x68`.
- Use the MPU6886 driver when the MPU6886 `WHO_AM_I` value is detected.
- Use the BMI270 driver when the BMI270 chip ID is detected.
- Normalize BMI270 accelerometer values to match the existing accelerometer unit expected by Core2 examples.
- Keep the existing `accelerometer` and `gyro` globals working for both IMU variants.
- Change the debugger baud rate from `921600` to `460800`. Core2 v1.0 used CP2104, but Core2 v1.1/v1.3 use CH9102, and `460800` is more reliable for the newer USB serial bridge.

Tested:
- Confirmed `examples/drivers/m5stack-imu` works on M5Stack Core2.
- Confirmed MPU6886-based Core2 and BMI270-based Core2 can be handled by the same device definition.

phoddie:
The video driver changes appear to be quite successful. Thank you for the additional effort. I need to make some time to read through the code.

phoddie:
I read through the code. I don't have any significant concerns with the changes. There is one thing I overlooked that I'd like to discuss.

The Moddable SDK renders 16-bit pixels in little-endian format. That's natural because the MCUs are little-endian, so doing the graphics calculations in little-endian is more efficient. The CO5300 display controller wants big-endian pixels. This driver swaps the pixels (using `__builtin_bswap16()`  in `co5300Send()`. That takes time - since it is a full read-write pass over all the pixels. 

We can't eliminate the swap in the driver (we would either need the IO layer to support byte swapping or the graphics engine to support big-endian rendering). But, we can swap more efficiently. In the 90 / 270 cases, we are making a separate full pass over the pixels to rotate them. That means there are two full passes through the pixels – one to swap and another to transform for rotation. If the swap is moved into the rotation transformation path, it should be faster, reducing the performance gap between rotated and unrotated cases. I fI had the hardware, I would try that myself. But, since I don't (yet!), would you kindly investigate that one last optimization? Then we can merge this. Thank you!

kitazaki:
@phoddie 

Thank you for reviewing the code and for the detailed explanation.
I will investigate whether the byte swap can be integrated into the rotation transformation path for the 90/270 degree cases.
It may take me a little time to check and test this on the hardware, but I will report back once I have confirmed the results.

kitazaki:
@phoddie 

Thank you for the suggestion. 
I moved the RGB565LE byte swap into the 90/270 degree software rotation path, so those cases no longer do a separate full pass over the pixels before rotating.

I tested this on hardware with the CO5300 rotation benchmark.
The 0 and 180 degree cases are unchanged, while the 90/270 full-width cases improved from about 20 ms/frame to 15 ms/frame. Narrow partial updates remain about 5 ms/frame.

Updated results:

| Rotation | Update area | Previous | Updated | Improvement |
|---|---:|---:|---:|---:|
| 0 | full width | 11 ms/frame | 11 ms/frame | no change |
| 180 | full width | 13 ms/frame | 13 ms/frame | no change |
| 90 | full width | 20 ms/frame | 15 ms/frame | 1.33x faster |
| 270 | full width | 20 ms/frame | 15 ms/frame | 1.33x faster |
| 90 | narrow partial | 5-6 ms/frame | 5 ms/frame | no change |
| 270 | narrow partial | 5 ms/frame | 5 ms/frame | no change |

phoddie:
You've gone from 40 ms for rotated frames to 15 ms. That's about 2.6x faster than the first attempt, and only 35% slower than unrotated. Very nice. Thank you for your persistence.

We'll get this merged.

mkellner:

phoddie:
I made a few changes in and around the driver after we merged the PR. The changes are in Moddable SDK 8.2.0.

- Poco now supports rendering in Big-Endian format. For the 0 and 180 cases, this completely eliminates the need to pass through the buffer to swap pixels, providing a nice speed bump. It also simplifies the 90 & 270 cases but not as much, since they still need to transform the buffer's orientation. Big-Endian rendering is enabled for `esp32/waveshare_amoled_206` now.
- In the unrotated case, the driver was setting the window for the next slice immediately after beginning asynchronous transmission of the pixels from the current slice. They effectively made all pixel transfers synchronous. The CO5300 driver borrowed this behavior from the ILI9341_P8 driver, but the important difference is that the ILI9341_P8 driver sends the window command asynchronously. I didn't want to restructure the driver to do the same, so I just moved the set window to before sending the slice. This results in at least a 10% frame rate boost in my tests when unrotated. Someone should check to see if the same needs to be done for 180.
- The code managing the brightness property was strangely complex and slightly buggy, so I simplified that.

Overall, it seems to be in better shape. I do still see some garbage in transitions when running conversationalAI. Those don't happen on Moddable Six or the simulator, so I'm reasonably sure the problem is in the driver. The issue predates the work to add rotation, so it is not a result of recent changes.

kitazaki:
@phoddie 

I tested the updated driver (Moddable SDK 8.2.0) and compared the results with my previous benchmark.

Rotation | Update area | Previous result | Updated driver | Result
-- | -- | -- | -- | --
0 | full width | 11 ms/frame | 8 ms/frame | 1.38x faster
180 | full width | 13 ms/frame | 10 ms/frame | 1.30x faster
90 | full width | 15 ms/frame | 15 ms/frame | no change
270 | full width | 15 ms/frame | 15 ms/frame | no change
90 | narrow partial | 5 ms/frame | 5 ms/frame | no change
270 | narrow partial | 5 ms/frame | 5 ms/frame | no change

The updated driver improves the 0-degree and 180-degree full-width update paths. 
The 90- and 270-degree full-width update paths remain at 15 ms/frame.

phoddie:
@kitazaki Thank you for rerunning the benchmarks. It is reassuring that the benchmarks confirm what I expected.

There's no reason for 180 to be slower than 0. I reviewed at the code today. `co5300SendHardwareRotated180()` is still behaving as-if it is transforming the pixels in software – it uses `co5300SendColorSync()`. It could use `esp_lcd_panel_io_tx_color()` to send the pixels asynchronously as in the 0 degree case. I think the right solution is to eliminate `co5300SendHardwareRotated180()` entirely and fold its functionality into the unrotated case – they are the same apart from the window calculation.

Do you want to try to do that? If not, I can try.

(Display drivers are so fundamental to performance of the UI that I hate to leave any performance optimization on the table.)

kitazaki:
@phoddie 

The 180-degree full-width update path has improved from 10 ms/frame to 8 ms/frame, and is now almost the same speed as the 0-degree path.

https://github.com/Moddable-OpenSource/moddable/pull/1623

phoddie:
I ran this on the hardware and was reminded that sometimes the updates have garbage. You can see that in piu/cards, for example, and contributed/conversationalAI. I eventually tracked that down.

Also, there was further optimization opportunity in the 90/270 cases by using asynchronous pixel writes. Finally, the 180 case was working but diverged from the unrotated case more than necessary.

All that will be fixed in the next release.

PR #1607, opened 2026-05-09
add Si12T touch driver

stc1988:
This driver is used to pet the [M5 StackChan](https://docs.m5stack.com/ja/StackChan). 　I don’t plan to add the StackChan device itself, but this driver might be needed elsewhere, so I created a PR.

[datasheet](https://m5stack-doc.oss-cn-shenzhen.aliyuncs.com/1205/Si12T_Datasheet_EN.pdf)

### DIFF
--- a/modules/drivers/sensors/si12t/si12t.js
+++ b/modules/drivers/sensors/si12t/si12t.js
@@ -1,6 +1,6 @@
 import Timer from "timer";
 
-const SENSITIVITY_REGISTERS = [0x02, 0x03, 0x04, 0x05, 0x06];
+const SENSITIVITY_REGISTERS = Uint8Array.of(0x02, 0x03, 0x04, 0x05, 0x06);
 const REG_CONFIG = 0x08;
 const REG_CONTROL = 0x09;
 const REG_REF_RESET1 = 0x0a;
@@ -11,8 +11,6 @@ const REG_CAL_HOLD1 = 0x0e;
 const REG_CAL_HOLD2 = 0x0f;
 const REG_OUTPUT1 = 0x10;
 
-const OUTPUT_NONE = 0;
-
 const TYPE_LOW = 0;
 const TYPE_HIGH = 1;
 
@@ -21,10 +19,10 @@ const LEVEL_7 = 7;
 const MIN_CHANNELS = 1;
 const MAX_CHANNELS = 12;
 
-const LOW_SENSITIVITY_VALUES = [0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77];
-const HIGH_SENSITIVITY_VALUES = [
-	0x88, 0x99, 0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff,
-];
+const LOW_SENSITIVITY_VALUES = Uint8Array.of(0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77);
+const HIGH_SENSITIVITY_VALUES = Uint8Array.of(
+	0x88, 0x99, 0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff
+);
 
 class Si12T {
 	#io;
@@ -32,7 +30,6 @@ class Si12T {
 	#sensitivityType = TYPE_LOW;
 	#sensitivityLevel = LEVEL_0;
 	#channels = 3;
-	#points;
 
 	constructor(options) {
 		const {
@@ -71,9 +68,20 @@ class Si12T {
 				});
 			}
 
-			this.#enableChannel();
-			this.#setControl2();
-			this.#setControl1();
+			// enableChannel
+			io.writeUint8(REG_REF_RESET1, 0x00);
+			io.writeUint8(REG_REF_RESET2, 0x00);
+			io.writeUint8(REG_CHANNEL_HOLD1, 0x00);
+			io.writeUint8(REG_CHANNEL_HOLD2, 0x00);
+			io.writeUint8(REG_CAL_HOLD1, 0x00);
+			io.writeUint8(REG_CAL_HOLD2, 0x00);
+			// setControl2
+			io.writeUint8(REG_CONTROL, 0x0f);
+			io.writeUint8(REG_CONTROL, 0x07);
+
+			// setControl1
+			io.writeUint8(REG_CONFIG, 0x22);
+
 			this.configure({
 				active: true,
 				sensitivityType,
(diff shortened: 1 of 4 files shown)


modules/drivers/sensors/si12t/si12t.js
```diff
@@ -0,0 +1,218 @@
...
+	 * - sensitivityType: 0 for low sensitivity range, 1 for high sensitivity range.
+	 * - sensitivityLevel: 0 through 7. Lower values detect smaller capacitance changes.
+	 * - channels: number of channels returned by sample(), 1 through 12. Defaults to 3.
+	 */
+	configure(options) {
+		if ("active" in options) {
+			const active = options.active;
+			this.#active = active;
+			if (active) this.#sleepDisable();
+			else this.#sleepEnable();
```
phoddie: #sleepDisable and #sleepEnable are only used once – just inline them here.

modules/drivers/sensors/si12t/si12t.js
```diff
@@ -0,0 +1,218 @@
...
+		return this.#points;
+	}
+
+	#readRegister(register) {
+		this.#assertOpen();
+		return this.#io.readUint8(register);
+	}
+
+	#writeRegister(register, value) {
+		this.#assertOpen();
```
phoddie: `#assertOpen()` isn't really necessary. If execution reaches here when closed, trying to call `this.#io.writeUint8` will throw anyway. All the `#assertOpen()` gives is a better error message in a rare case.

modules/drivers/sensors/si12t/si12t.js
```diff
@@ -0,0 +1,218 @@
...
+		};
+	}
+
+	sample() {
+		const count = Math.ceil(this.#channels / 4);
+
+		for (let i = 0; i < count; i++)
+			this.#parseTouchByte(this.#readRegister(REG_OUTPUT1 + i), i * 4);
+
+		return this.#points;
```
phoddie: `sample()` should return a fresh object each time. If the application collects samples, this code won't work as expected:

```js
const samples = [];
for (let i = 0; i <= 10; I++)
     samples.push(sensor.sample());
```

It may be slightly faster to continue using `this.#points` internally and then do `return this.#points.slice()` rather than starting with a new array each time.

modules/drivers/sensors/si12t/si12t.js
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	#readRegister(register) {
+		this.#assertOpen();
+		return this.#io.readUint8(register);
+	}
+
+	#writeRegister(register, value) {
+		this.#assertOpen();
+		this.#io.writeUint8(register, value & 0xff);
```
phoddie: The `& 0xff` is unnecessary. Only the low 8 bits of the number are used.

modules/drivers/sensors/si12t/si12t.js
```diff
@@ -0,0 +1,218 @@
...
+const LEVEL_0 = 0;
+const LEVEL_7 = 7;
+const MIN_CHANNELS = 1;
+const MAX_CHANNELS = 12;
+
+const LOW_SENSITIVITY_VALUES = [0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77];
+const HIGH_SENSITIVITY_VALUES = [
+	0x88, 0x99, 0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff,
+];
+
```
phoddie, line 26: I think you want to freeze the static arrays so they live entirely in a flash. An easy way to do that is

```js
Object.freeze({SENSITIVITY_REGISTERS, LOW_SENSITIVITY_VALUES, HIGH_SENSITIVITY_VALUES}, true);
```

Also, since the values are all 8-bit, you reduce their footprint using `Uint8Array`:
```js
const LOW_SENSITIVITY_VALUES = Uint8Array.of(0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77);
```

phoddie (commented):
Generally this looks good. It kind of feels like it was translated from C? There are a lot of single use private methods. I marked one place that happens, but there are more. II suggest eliminating those in most cases:
- There is some call overhead
- There is some code overhead
- There is RAM overhead – because each private method uses a RAM slot as JavaScript requires them to be implemented as private fields.

stc1988:
> Generally this looks good. It kind of feels like it was translated from C? There are a lot of single use private methods. I marked one place that happens, but there are more. II suggest eliminating those in most cases:
> 
> There is some call overhead
> There is some code overhead
> There is RAM overhead – because each private method uses a RAM slot as JavaScript requires them to be implemented as private fields.

Thanks for the review. You’re right—I did refer to a C implementation, but I also adjusted some parts, like making certain logic more explicit so it works better as comments for Moddable. Still, it looks like there’s more I need to improve, so I’ll take another look.

phoddie:
Thanks for being open to taking another look. I find that small, relatively simple projects like this one are a good opportunity to refine how I organize code.

> ...like making certain logic more explicit so it works better as comments for Moddable...

Interesting. I sensed that the helper function names were kind of serving the purpose of comments. As many people have observed, I'm not a big user of actual comments in code, but they do have zero runtime cost. ;)

modules/drivers/sensors/si12t/si12t.js
stc1988: I removed the function and updated it to handle the processing directly within configure.

modules/drivers/sensors/si12t/si12t.js
stc1988: I removed the unnecessary `& 0xf`.

modules/drivers/sensors/si12t/si12t.js
stc1988: I revised the implementation so that it returns a new object each time.

modules/drivers/sensors/si12t/si12t.js
stc1988: I used the latter approach and froze the static array.

modules/drivers/sensors/si12t/si12t.js
stc1988: I removed necessary `#assertOpen()`.

stc1988:
I reviewed and removed unnecessary private functions across the entire codebase, not just in the areas that were pointed out. Of course, I’ll also review everything myself, but I’ll incorporate these review learnings into the AI as well.

phoddie:
Thank you for the changes. This is more straightforward to read and has a lighter runtime impact too. Best of both worlds.

mkellner:

PR #1608, opened 2026-05-09
fix(mcpack): replicate external subplatform support from mcmanifest

HipsterBrown:
External subplatform (a.k.a custom device targets) support was added in Moddable v8.1.0 but only in mcmanifest.js for `mcconfig` to resolve. This PR creates parity support for `mcpack` as well.

### DIFF
--- a/tools/mcpack.js
+++ b/tools/mcpack.js
@@ -171,16 +171,40 @@ export default class extends TOOL {
 				if (this.platform)
 					throw new Error("-p '" + name + "': too many platforms!");
 				name = name.toLowerCase();
-				let parts = name.split("/");
+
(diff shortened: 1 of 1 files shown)


phoddie:
Thanks for catching this and for making the change to keep them in sync. 

I noticed recently that xs-dev defaults to mcpack for new projects. That's a great way to make sure mcpack gets exercised.

HipsterBrown:
Yea, my approach for xs-dev these days is to be a bridge for JS developers to the embedded world by providing a familiar workflow, I.e. package.json as the primary manifest, the polyfilled APIs provided by mcpack, and ECMA-419 first. 

It will always support manifest.json and all the ways Moddable projects that can be built.

mkellner:

PR #1609, opened 2026-05-10
M5StickC and M5StickCplus migrate drives from pins/i2c to Ecma-419

stc1988:
Migrate below drivers from pins/i2c to Ecma-419 sensor and peripheral.
- AXP192
- BM8563
- MPU6886
- SH200Q(M5StickC only)

And I updated the M5StickC-related examples to use these drivers.

### DIFF
--- a/build/devices/esp32/targets/m5stick_c/host/provider.js
+++ b/build/devices/esp32/targets/m5stick_c/host/provider.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021  Moddable Tech, Inc.
+ * Copyright (c) 2021-2026  Moddable Tech, Inc. ,Satoshi Tanaka
  *
  *   This file is part of the Moddable SDK Runtime.
  *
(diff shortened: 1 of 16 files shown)


stc1988:
There’s an additional update.

I enabled the PEK (Power Enable Key) on the AXP192 so that short presses of the power button can be detected. However, since pekstate needs to be polled, it’s disabled by default.

phoddie:
The changes look good. The power button implementation is a reasonable solution given that polling is necessary.

For future reference, you can now write these...

```js
const white = render.makeColor(255, 255, 255);
const blue = render.makeColor(0, 0, 255);
```

as 

```js
const white = render.makeColor("white");
const blue = render.makeColor("blue");
```

Since they are cached at start-up, there's no meaningful performance difference. The readability is nice. Also, the list of colors can be extended by the host. That means that a Stack-chan host could provide color names that are used in the Stack-chan UI. I suggest sticking to color names (e.g. "bright orange" and not drift into uses ("mouth", "background", etc).

mkellner:

PR #1610, opened 2026-05-12
add IMU sensor to ws_round

stc1988:
This PR adds IMU sensor to `pico/ws_round/` and `pico/ws_rounc_touch` and QMI8658 examples.


mkellner:
I tweaked the qmi8658/bounce demo to use the qmi8658 sensor and committed.

PR #1611, opened 2026-05-16
Implement Promise job scheduling for WASM

meganetaaan:
This PR implements Promise job scheduling for the WASM XS platform.

The WASM platform previously had an empty `fxQueuePromiseJobs()` implementation, so queued Promise jobs were not scheduled back onto the browser/event-loop side. As a result, `await` continuations and other Promise-driven callbacks could remain pending when running XS applications compiled to WebAssembly.

This became visible while bringing up the [Stack-chan WASM simulator](https://stack-chan.github.io/stack-chan/web/simulator/): the application could reach the WASM launch path, but async application hooks and Promise-based UI/runtime callbacks did not reliably continue. In practice, this made symptoms look like the simulator or Piu UI was stuck even though the WASM module itself had loaded.

## What changed

- Add a WASM-specific Promise job timer using Emscripten timers.
- Schedule `fxRunPromiseJobs()` with `emscripten_set_timeout(..., 0, ...)` when Promise jobs are queued.
- Avoid scheduling duplicate timers while one is already pending.
- Clear the pending timer when the machine platform is deleted.
- Store the timer handle in the WASM machine platform state.

## Details

The implementation mirrors the intent of other platform-specific `fxQueuePromiseJobs()` implementations: queued Promise jobs need to be dispatched asynchronously by the host/platform event loop.

For WASM, this PR uses Emscripten's timeout API to enqueue a zero-delay callback. The callback clears the stored timer handle and then runs the pending Promise jobs with `fxRunPromiseJobs(the)`.

## Motivation / observed issue

Without this implementation, Promise continuations in WASM builds may not progress after being queued. For applications that rely on `async`/`await`, this can prevent code after an `await` from running.

In the Stack-chan WASM simulator work, this affected async hooks and Promise-driven simulator/Piu behavior. After applying this patch, Promise continuations are scheduled through the WASM event loop, allowing async application code to resume as expected.

## Files changed

- `xs/platforms/wasm_xs.c`
- `xs/platforms/wasm_xs.h`


phoddie:
Thanks for the PR. The change makes perfect sense. We'll get this merged.

mkellner:

PR #1614, opened 2026-05-23
Fix building on NixOS

alesya-h:
This makes it possible to build moddable on NixOS by removing hardcoded paths. Without it I couldn't build PebbleOS with moddable support on my system. I have sent signed CLA.


mkellner:

PR #1619, opened 2026-05-29
Add 'info instruments' help entry to xsdb

stc1988:
Recently, in development, I’ve been trying to have an AI agent use xsdb to verify its behavior. Since I’m instructing it to look up how to use xsdb on its own via the help command, writing good help documentation is important.

### DIFF
--- a/tools/xsbug-log/xsbug-debugmachine.js
+++ b/tools/xsbug-log/xsbug-debugmachine.js
@@ -2844,7 +2844,8 @@ class DebugMachine extends Machine {
 						{ keys: "eval <expr>", desc: "Evaluate JavaScript expression in current frame" },
 						{ keys: "info modules", desc: "List loaded modules" },
 						{ keys: "info module name", desc: "Show exports" },
-						{ keys: "info modules.name.path", desc: "Navigate into module exports" }
+						{ keys: "info modules.name.path", desc: "Navigate into module exports" },
+						{ keys: "info instruments", desc: "Print instrument snapshots" }
 					]
 				},
 				{


phoddie:
Good catch. We'll merge that.

FWIW – I think one reason this wasn't documented is that is just a primitive first attempt at instrumentation support in xsdb. I have always plans to enhance the instrumentation capabilities of xsdb – but haven't quite sorted out how to do that.

> Recently, in development, I’ve been trying to have an AI agent use xsdb to verify its behavior

Very cool. I've had some interesting success with that. Good instructions are important. I hope you'll share your experiences with that at some point.

stc1988:
Basically, I write instructions like the following in the documentation.

```
- The `-dl` option starts `xsdb`, a debugger that can be used similarly to `gdb`. Use the `help` command inside `xsdb` for command details.
- Use `xsdb` to confirm the app starts without errors and emits the expected logs.
- When verification is complete, exit `xsdb` with `quit`.
```

Previously, the focus was mostly on verifying builds with `mcconfig -t build`, but by using `xsdb`, we can now confirm that it at least starts up without errors, as in `test-examples`. It also uses the logs to determine whether the expected processing sequence was executed.

This time, I had it implement preload support by comparing the instruments before and after the change, and it seems to be working well.

phoddie:
Thanks for the notes.

Checking the impact of preload is a nice use of the instrumentation in xsdb. 

I imagine that test-example could be extended to do basic verification of behavior – such as confirming that certain logs are received, for example. That would require some specification information for the app, but would give a deeper validation.

One thing I have found that the LLMs struggle a bit with is initially communicating with xsdb. They over have to try several approaches to get to the point where they can use xsdb interactively. Have you found instructions that bypass that? Or maybe you are just luckier? ;)

stc1988:
I’m using Codex, and it’s been working smoothly.
It might be because it remembers my past trial and error, so I don’t need to give very explicit instructions.

phoddie:
>  I’m using Codex, and it’s been working smoothly. It might be because it remembers my past trial and error, so I don’t need to give very explicit instructions.

I would be great if each developer didn't have to rediscover how to interface xsdb to the agent harness. Maybe you could prompt Codex to summarize its technique into a markdown document? I'd be interested to see what it says and how we might make that knowledge more available / discoverable. Thank you for your consideration.

PR #1623, opened 2026-06-07
Fix CO5300 updates when using 180-degree rotation

kitazaki:
## Summary

This change removes the separate synchronous color-transfer path for `rotation == 2` and routes it through the same color send flow used by the normal orientation. For 180-degree rotation, the update window is  still adjusted per chunk based on the remaining update lines.

Rotation | Moddable SDK 8.2.0 | Updated driver | Result
-- | -- | -- | --
0 | 8 ms/frame | 8 ms/frame | no change
180 | 10 ms/frame | 8 ms/frame | 1.25x faster
90 full width | 15 ms/frame | 15 ms/frame | no change
270 full width | 15 ms/frame | 15 ms/frame | no change
90 narrow | 5 ms/frame | 5 ms/frame | no change
270 narrow | 5 ms/frame | 5 ms/frame | no change

## Changes

  - Set the CO5300 update window per transferred chunk for `rotation == 2`
  - Reuse the existing async color transfer / queue handling in `co5300Send`
  - Remove the dedicated `co5300SendHardwareRotated180` helper


mkellner:

PR #1627, opened 2026-06-09
fix base64 decode heap overflow on stray padding

alhudz:
Repro: `Base64.decode("AAAA=")` (or `"AAAAAAAA=="`).
Expected: 3 decoded bytes from the `AAAA` quantum, stray `=` ignored.
Actual: heap-buffer-overflow. `xs_base64_decode` sizes the buffer as `(srcSize / 4) * 3` then decrements once per trailing `=`, but the decode loop skips a `=` that lands on a quantum boundary (`srcIndex` 0 or 1) instead of treating it as padding, so it still emits 3 bytes for each preceding full group. Allocation is 2, three bytes get written.
Cause: the padding decrements assume every `=` shortens the last group, which is false for stray padding, so the estimate falls below the bytes actually produced.
Fix: drop the decrements so the allocation stays a true upper bound; the existing `xsmcSetArrayBufferLength` call already trims the result to the exact decoded length, so valid padded input is unaffected. The XS core (`fxUint8ArrayFromBase64`) and `JSONBase64Parser` already round up rather than subtract.

Confirmed under ASan with a standalone copy of the decode loop: pre-patch `AAAA=` allocates 2 and writes 3 (`WRITE of size 1 ... 0 bytes after 2-byte region`); post-patch writes never exceed the allocation and padded inputs keep their lengths (`QQ==` -> 1, `QUJDRA==` -> 4).

### DIFF
--- a/modules/data/base64/modBase64.c
+++ b/modules/data/base64/modBase64.c
@@ -109,10 +109,6 @@ void xs_base64_decode(xsMachine *the)
 	srcSize = c_strlen((char *)src);
 
 	dstSize = (srcSize / 4) * 3;
-	if (c_read8(src + srcSize - 1) == '=')
-		dstSize--;
-	if (c_read8(src + srcSize - 2) == '=')
-		dstSize--;
 	srcIndex = 0;
 
 	xsmcSetArrayBufferResizable(xsResult, NULL, dstSize, dstSize);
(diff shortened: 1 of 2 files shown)


phoddie:
Thank you for the report and patch. The change looks reasonable, leveraging the resizable ArrayBuffer. The original implementation predates resizable ArrayBuffer support in JavaScript and so needed a precise calculation of the buffer size for the initial allocation, but this is no longer the case.

Note that this module is [deprecated](https://github.com/Moddable-OpenSource/moddable/blob/52fe713a1d4d0aed403c6307e87d0559941bb3af/modules/data/base64/manifest.json#L13) in favor of the Base64 support that is now standard in JavaScript.

To merge this PR, we require a completed [Contributor License Agreement](https://github.com/Moddable-OpenSource/moddable/tree/public/licenses#contributor-license-agreement). That can be emailed to Moddable at the address provided.

Finally, this issue almost certainly was not found by runtime inspection tools (e.g. ASAN). Even most static analyzers would not recognize `xsmcSetArrayBufferResizable` as an allocator and so would fail to detect the overflow. So, I'm curious how you uncovered this issue. Would you kindly share?

alhudz:
I have completed and emailed the signed Contributor License Agreement (CLA).
Thanks

phoddie:
We've received the CLA. Thank you. We'll get this merged for our next Moddable SDK release.

mkellner:

alhudz:
Thanks! Appreciated

PR #1630, opened 2026-06-10
Fix request header forwarding in pebbleproxy

jplexer:
Empty header lines and colons in header values broke every fetch(). Skip empty lines, split on the first colon only, and emit "\n" between headers instead of after each.

### DIFF
--- a/build/devices/pebble/modules/httpclient/httpclient-pebble.js
+++ b/build/devices/pebble/modules/httpclient/httpclient-pebble.js
@@ -235,7 +235,8 @@ class HTTPClient {
 							current.sending = parseInt(item.value[1]);
 						else if (("transfer-encoding" === item.value[0]) && ("chunked" === item.value[1].toLowerCase()))
 							current.sending = true;		// unknown length
-						this.#remain = ArrayBuffer.fromString(`${item.value[0]}:${item.value[1]}\n`);
+						this.#remain = ArrayBuffer.fromString(`${current.sentHeader ? "\n" : ""}${item.value[0]}:${item.value[1]}`);
+						current.sentHeader = true;
 						this.#remain = new Uint8Array(this.#remain);
 						this.#remain.part = 3;
 						this.#remain.position = 0;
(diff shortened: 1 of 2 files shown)


phoddie:
Please share a test case to reproduce this problem. Thank you.

build/devices/pebble/modules/httpclient/httpclient-pebble.js
```diff
@@ -235,7 +235,8 @@ class HTTPClient {
 							current.sending = parseInt(item.value[1]);
 						else if (("transfer-encoding" === item.value[0]) && ("chunked" === item.value[1].toLowerCase()))
 							current.sending = true;		// unknown length
-						this.#remain = ArrayBuffer.fromString(`${item.value[0]}:${item.value[1]}\n`);
+						this.#remain = ArrayBuffer.fromString(`${current.sentHeader ? "\n" : ""}${item.value[0]}:${item.value[1]}`);
+						current.sentHeader = true;
```
phoddie, line 239: The way I read  this, only the first header will be sent, all others will be replaced with `\n`. Maybe this change is unnecessary given the change to the proxy which ignores lines without a colon, which will skip the empty line at the end of the headers?

build/devices/pebble/npm/pebbleproxy/proxy.js
```diff
@@ -173,8 +173,10 @@ function httpMessage(id, e) {
 			request.xhr.responseType = 'arraybuffer';
 
 			request.headers.split("\n").forEach(line => {
-				const [key, value] = line.split(":");
-				request.xhr.setRequestHeader(key, value);
+				const colon = line.indexOf(":");
+				if (colon <= 0)
+					return;		// skip empty lines, including the serializer's trailing newline
+				request.xhr.setRequestHeader(line.slice(0, colon), line.slice(colon + 1));
```
phoddie, line 179: This looks good. I didn't consider that the header value could have a colon....

phoddie:
I've merged the proxy.js changes. If the httpclient-pebble.js changes are necessary, please let me know.

PR #1636, opened 2026-06-13
Add GDEY037T03 e-paper driver

jplexer:
In my spare time I've been working on writing custom firmware for a commercial bedside clock by reversing the circuit. The only driver missing was for the Display (GDEWY037T03) and I thought if I wrote it anyway might as well upstream it :)

### DIFF
--- a/modules/drivers/gdey037t03/gdey037t03.c
+++ b/modules/drivers/gdey037t03/gdey037t03.c
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2026 Joshua Jun
+ *
+ *   This file is part of the Moddable SDK Runtime.
+ *
+ *   The Moddable SDK Runtime is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU Lesser General Public License as published by
+ *   the Free Software Foundation, either version 3 of the License, or
+ *   (at your option) any later version.
+ *
+ *   The Moddable SDK Runtime is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU Lesser General Public License for more details.
+ *
+ *   You should have received a copy of the GNU Lesser General Public License
+ *   along with the Moddable SDK Runtime.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "xsmc.h"
+#include "xsHost.h"
+
+// Geometry must match gdey037t03.js: logical 416 x 240, native 240 x 416.
+#define STRIDE 52			// logical row bytes (416 >> 3)
+#define NATIVE_STRIDE 30	// native row bytes (240 >> 3)
+#define NATIVE_H 416
+
+// transpose(logicalBuffer, outBuffer): native(nx, ny) = logical(ny, nx).
+// Each native output byte packs 8 logical columns at one native row.
+void xs_gdey037t03_transpose(xsMachine *the)
+{
+	uint8_t *logical, *out;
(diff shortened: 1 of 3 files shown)


phoddie:
Super cool.

It looks perfect.

I only had time for a quick read this afternoon. A couple of notes:

- Freezing the prototype is unnecessary as `xsl` (the XS linker) takes care of that. It didn't always, which is why there's still some SDK code doing it.
- The only performance bottleneck appears to be `#toNative()`, but it probably doesn't matter for low-frequency display updates. The cycle counter in me would process 8 pixels at a time in the inner loop to reduce the calculations to write to the output buffer.

> ...writing custom firmware for a commercial bedside clock...

This does lead to a couple of natural questions... what's the product, and how can I try this? ;)

jplexer:
> This does lead to a couple of natural questions... what's the product, and how can I try this? ;)

The product is the [Habity](https://habity.design) and i've been working on this thing called [openhabit](https://github.com/jplexer/openhabit)

modules/drivers/gdey037t03/gdey037t03.js
```diff
@@ -0,0 +1,307 @@
...
+const HEIGHT = 240;
+const STRIDE = WIDTH >> 3;
+const NATIVE_W = HEIGHT;
+const NATIVE_H = WIDTH;
+const NATIVE_STRIDE = NATIVE_W >> 3;
+const FRAME_BYTES = STRIDE * HEIGHT;
+
+// Native transpose (gdey037t03.c): native(nx, ny) = logical(ny, nx). Done in C
+// because the per-pixel bit loop over the whole frame is far too slow in script.
+function transpose(logical, out) @ "xs_gdey037t03_transpose";
```
phoddie, line 44: The `@` syntax for declaring native functions is deprecated in favor of the newer `native` approach. Here that would be:

```js
function transpose(logical, out) { return native("xs_gdey037t03_transpose").call(this, logical, out); }
```

See the [XS in C](https://www.moddable.com/documentation/xs/XS%20in%20C#native) docs for details.

modules/drivers/gdey037t03/gdey037t03.c
```diff
@@ -0,0 +1,59 @@
...
+#define NATIVE_H 416
+
+// transpose(logicalBuffer, outBuffer): native(nx, ny) = logical(ny, nx).
+// Each native output byte packs 8 logical columns at one native row.
+void xs_gdey037t03_transpose(xsMachine *the)
+{
+	uint8_t *logical, *out;
+	xsUnsignedValue logicalLen, outLen;
+
+	xsmcGetBufferReadable(xsArg(0), (void **)&logical, &logicalLen);
```
phoddie, line 36: You could eliminate the `#toNative()` wrapper by allocating the buffer here using XS in C and returning it. This allocates a buffer the same length as the input buffer.

```c
xsmcSetInteger(xsResult, (int)logicalLen);
xsResult = xsNew1(xsGlobal, xsID_Uint8Array, xsResult);
xsmcGetBufferWritable(xsResult, (void **)&out, &outLen);
```

phoddie (changes requested):
The C version of the screen buffer transform will be faster. ;)

I added a couple of small comments. Let you don't want to apply them, that's OK. Just let me know either way and we'll get this merged.

phoddie:
Also, the clock looks very nice and your firmware is very straightforward. Is a soldering iron required to flash the firmware?

jplexer:
Thanks for your comments on the code Peter! I'll take a look tomorrow.



> Also, the clock looks very nice and your firmware is very straightforward. Is a soldering iron required to flash the firmware?

Nope! Don't even need to open it up, you can just flash it using the USB-C port and ESP-IDF (using mcconfig ;D)

mkellner:

PR #1639, opened 2026-06-16
reject png bit depths that overflow the scanline slop

rootvector2:
`xs_PNG_constructor` never validates the IHDR bit depth, so a png with more than 8 bits per channel makes `filterBytesPerPixel` exceed the fixed `kScanLineSlop` and the leftmost-pixel `prev = pix - bpp` read in the `subFilter`/`paethFilter` unfilter step runs before the scanline allocation, found with an asan build decoding a 16-bit rgba png.

### DIFF
--- a/modules/commodetto/commodettoReadPNG.c
+++ b/modules/commodetto/commodettoReadPNG.c
@@ -252,6 +252,8 @@ void xs_PNG_constructor(xsMachine *the)
 				png->filterBytesPerPixel = (bitsPerPixel + 7) >> 3;
 
 				//@@ check for supported combinations
+				if ((0 == png->channelCount) || (png->bitDepth > 8) || (0 == png->bitDepth))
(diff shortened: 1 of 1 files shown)


phoddie:
I think it would be more explicit to validate the parameters that contribute to `filterBytesPerPixel` rather than the resulting value:

```c
				if((0 == png->channelCount) || (png->bitDepth > 8) || (0 == png->bitDepth))
					xsErrorPrintf("unsupported");
```

This also halts decoding when an invalid `channelCount` is encountered. Does that still work with your tests?

rootvector2:
makes sense, switched to validating `channelCount` and `bitDepth` directly. since `png` is `c_calloc`'d an invalid `colorType` leaves `channelCount` at 0, so that branch catches it too. retested: 16-bit rgba still rejected before the unfilter read, 8-bit still decodes clean.

mkellner:

PR #1624, opened 2026-06-07
M5Stack Core2 Use ECMA-419 drivers and Add power button support

stc1988:
### M5StackCore2
- Use ECMA-419 drivers(touch, RTC, power)
- Optional power button support
- add M5StackCore2 to esp32.md document
- Refactor Host provider

M5StackCore2 has several versions which has different sensors and peripherals.

Version | IMU | Power
-- | -- | --
[v1 ](https://docs.m5stack.com/ja/core/core2)| MPU6886 | AXP192
[v1.1](https://docs.m5stack.com/ja/core/Core2%20v1.1) | MPU6886 | AXP2101
[v1.3](https://docs.m5stack.com/ja/core/Core2_v1.3) | BMI270 | AXP192

I refactored the code to separate it into different files because trying to handle these differences in the host provider made the code harder to follow.

### M5StackCores3
- Optional power button support

PR #1642, opened 2026-06-17
Revert the removal of startupSound and startupVibration.

stc1988:
In #1624, I accidentally committed the settings I used for testing, so I’m reverting them.


mkellner:

PR #1643, opened 2026-06-19
bound BMF block walk to buffer end

rootvector2:
the BMF block walk in `xs_parseBMF` and `CFESetFontData` advances `bytes` by file-controlled block sizes (`bytes += 4 + c_read32(bytes)`, `bytes += size - 8`) without bounding the cursor against the buffer end, so a crafted block size makes the next `c_read*` read past the allocation (found with an asan build feeding a truncated font); this bounds every advance and field read against `end`.

### DIFF
--- a/modules/commodetto/cfeBMF.c
+++ b/modules/commodetto/cfeBMF.c
@@ -77,7 +77,7 @@ void CFELockCache(CommodettoFontEngine bmf, uint8_t lock)
 
 void CFESetFontData(CommodettoFontEngine bmf, const void *fontData, uint32_t fontDataSize)
 {
-	const unsigned char *bytes, *start;
+	const unsigned char *bytes, *start, *end;
 	uint32_t size;
 	uint8_t version;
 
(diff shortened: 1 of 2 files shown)


phoddie:
Thanks for your patience. I'm just looking into a couple scenarios.

phoddie:
I did [a pass through](https://gist.github.com/phoddie/5cdfaf9fd53ac008eaa504ffcc3fc5d6) cfeBMF.c. Some of the changes are just clean-up, others (kerning table) are functional.  Please take a look, as I don't have access to your test harness. Maybe you can apply any relevant changes to the other file as well? Thanks!

rootvector2:
ran your cfeBMF pass through the asan harness. one thing stood out: the common block's `pages` field is at offset 8, but the gist reads it at `bytes + 6`, which lands on `scaleH` (texture height, basically never 1), so every valid font got rejected. kept that read at `bytes + 8` and the valid font parses again (charCount=1, height=20, ascent=16, kernCount=1). the rest holds up, including requiring `size >= 20` so `charTable[19]` is always in bounds, and the `>= 11` kern guard.

applied the relevant bit to `commodettoParseBMF.c` too: it only checked `size % 20` on the chars block, so i added the same `size > (end - bytes) - 4` bound there. crafted and truncated inputs reject cleanly, no asan hits.

phoddie:

rootvector2:
thanks for picking this up, saw it land on `public`. the kern-size bound you added in `tighten & sync` covers a case my `>= 11` guard missed.

phoddie:
Thanks for the help!

Issue #1655, opened 2026-07-13
ESP32: use-after-free panic in ECMA-419 TCP tcpReceive when a socket closes with data in flight (lwIP tcpip thread vs XS thread race)


xymeow:
## Environment

- Moddable SDK 8.2.3 (the relevant code in `modules/io/socket/lwip/tcp.c` is unchanged on the current `public` branch)
- ESP-IDF 6.0, ESP32-S3 (M5Stack CoreS3), `CONFIG_LWIP_TCPIP_CORE_LOCKING` not set (IDF default)
- Debug build (`mcconfig -d`); the device runs an HTTP server (`embedded:io` listener) serving roughly 0.2–1 request/s from a LAN client, plus occasional outbound HTTP (WAV streaming)

## Symptom

Sporadic `Guru Meditation Error: Core 1 panic'ed (LoadProhibited)` followed by `rst:0xc (RTC_SW_CPU_RST)`. 24 such resets were captured on the serial console over one evening of ordinary traffic. Three panics include full backtraces, all entering through `tcpReceive` on the lwIP `tcpip_thread`:

```
tcpReceive               modules/io/socket/lwip/tcp.c:575
tcp_process_refused_data lwip/src/core/tcp.c:1578
tcp_input                lwip/src/core/tcp_in.c:434
ip4_input / ethernet_input / tcpip_thread
```

```
tcpReceive               modules/io/socket/lwip/tcp.c:575
tcp_input                lwip/src/core/tcp_in.c:512
ip4_input ...
```

```
fxInNetworkDebugLoop     xs/platforms/esp/xsPlatform.c:122
tcpReceive               modules/io/socket/lwip/tcp.c:559
...
```

## Analysis

`tcpReceive` runs on the lwIP `tcpip_thread`. `doClose` / `xs_tcp_destructor` run on the XS task and clear the lwIP callbacks with raw `tcp_recv/tcp_sent/tcp_err` calls and then free the `TCP` record. (`tcp_close` itself *is* marshaled on ESP32 via `tcp_close_safe`/`tcpip_api_call` — the callback removal and the free are the unsynchronized steps.) With core locking disabled there is no synchronization between the two threads, so:

[...]

PR #1656, opened 2026-07-13
ESP32 419 TCP: marshal the teardown callback clear to the tcpip task

xymeow:
Addresses the socket-teardown race in #1655, per @phoddie's `TCP_EVENT_RECV` analysis there.

lwip reads `pcb->recv` and then `pcb->callback_arg` at separate instants, so clearing the callbacks from the XS task can interleave with a dispatch already in progress on the tcpip task. This routes the XS-side teardown (`doClose`, `xs_tcp_destructor`) through a marshaled `tcp_clear_callbacks_safe`, so the clears run on the tcpip task and no callback can be in flight when it returns. `removeTCPCallbacks` stays for the callers that already run on the tcpip task (`tcpError`, the `tcpReceive` error path), where plain stores are safe. On single-threaded (`!ESP32`) builds `tcp_clear_callbacks_safe` is a plain macro.

Verified on ESP32-S3 (M5Stack CoreS3), ESP-IDF 6.0: no functional regressions and clean through heavy accept/close churn (including RST-aborted connections).

**Scope:** this is the teardown fix only, which is the part we converged on. It does **not** claim to close the separate `tcp->buffers` receive-buffer corruption still discussed in #1655 — that fault (a receive-buffer node freed under the walk, always `EXCVADDR = 0xe2f61b44`) persists even with full `builtinCriticalSection` protection of the list, so it looks like a distinct memory issue and is left open there.

CLA on file (emailed to info@moddable.com).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/io/socket/lwip/tcp.c
+++ b/modules/io/socket/lwip/tcp.c
@@ -239,7 +239,12 @@ void xs_tcp_destructor(void *data)
 	if (!tcp) return;
 
 	if (tcp->skt) {
-		removeTCPCallbacks(tcp);
+		// Marshaled: TCP_EVENT_RECV reads pcb->recv and pcb->callback_arg at
+		// separate instants, so clearing the callbacks from the XS task can
+		// interleave with a dispatch already in progress on the tcpip task.
+		// Running the clears on the tcpip task itself closes that window.
+		tcp_clear_callbacks_safe(tcp->skt);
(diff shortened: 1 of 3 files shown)


phoddie:
Cool, thanks. This looks clean and matches our discussion. We'll get this merged.

mkellner:

PR #1657, opened 2026-07-19
Fix Linux Piu wheel scrolling

meganetaaan:
## Summary

Fix mouse wheel scrolling in Linux Piu applications.

Linux GTK scroll events produced scrolling that was both reversed and much smaller than on other platforms. This change normalizes the direction and amount to match the behavior of the Windows and other Piu backends.

## Testing

- Built `xsbug` for `x-lin`.
- Built `mcsim` for `x-lin`.
- Verified wheel scrolling direction and speed in `xsbug` on Sway/Wayland.

### DIFF
--- a/modules/piu/PC/lin/piuView.c
+++ b/modules/piu/PC/lin/piuView.c
@@ -20,6 +20,10 @@
 
 #include "piuPC.h"
 
+#define piuWheelDelta 120
+/* GTK reports a standard Wayland wheel notch as 1.5 smooth scroll units. */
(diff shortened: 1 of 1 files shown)


phoddie:
Nice. There is so much variability in Linux, I wonder if this change will work well across distributions. There's only one practical way to find out.... merging it. ;) So, we'll do that.

Thank you!

mkellner:
This is merged and will be available in the next release. Thank you.

PR #1658, opened 2026-07-20
add device esp32/c3_touch_1_28

stc1988:


phoddie:
Thanks for the PR. The device support looks straightforward. Too bad the hardware doesn't include a knob (Pulse Count) so it can be turned.

I found that the PDF the documentation gives a 404 error. Would you double check that?
> https://docs.ai-thinker.com/_media/esp32/docs/esp-c3-32s-kit-v1.0_specification.pdf

stc1988:
> Thanks for the PR. The device support looks straightforward. Too bad the hardware doesn't include a knob (Pulse Count) so it can be turned.

A knob definitely sounds interesting. I'll look for a compatible device next time.

> I found that the PDF the documentation gives a 404 error. Would you double check that?
https://docs.ai-thinker.com/_media/esp32/docs/esp-c3-32s-kit-v1.0_specification.pdf

The documentation has been updated since I found a different working link.

phoddie:
> A knob definitely sounds interesting. I'll look for a compatible device next time.

That would be cool.

mkellner:
The esp32-c3 does not support the hardware Pulse Counter (PCNT).
There are s3 versions that contain a knob.

mkellner:

PR #1676, opened 2026-08-10
add esp32/m5nanoh2

stc1988:
I revised the setup implementation for the M5NanoC6 to match the M5NanoH2.


phoddie:
Thanks for the PR. It generally looks fine.

There is one change that is unusual, the modification of [esp32/manifest.json](https://github.com/stc1988/moddable/blob/d65eebc563a2e8cbdfc663772377f2403b9e2c85/modules/io/manifests/esp32/manifest.json#L124-L127). This is the only time an M5 device has been added to this file. I acknowledge that the use of the esp32/manifest.json is a bit inconsistent.

If possible it would be best to keep support for individual boards in their own manifest rather than the overarching ESP-IDF manifest. Would you see if that is possible here? Thank you.

stc1988:
I modified esp32/manifest.json because a build error occurred when including io/manifest.json on this board.
Since esp32h2 does not have Wi-Fi, it is necessary to avoid including manifest_net.json.
Currently, I cannot think of a way to handle this on the individual board side.

phoddie:
Ah, thanks. I understand now.

Maybe it is a mistake the ESP32 targets include `manifest_net.json`? Even for silicon with Wi-Fi, there are plenty of projects that don't use it and so don't need to carry it. That change will have a lot of ripple effects though...

stc1988:
Yes, I did feel it was a bit odd to include manifest_net.json, since it increases the build size for projects that do not use Wi-Fi.

I understand that developers are supposed to include manifest_net.json in projects that use Wi-Fi, so I think it would be fine to exclude it from the ESP32 targets. However, since the scope of impact is indeed large, I went ahead with this fix for now.

phoddie:
I've made changes to eliminate the inclusion of manifest_net.json by platform manifests. That should eliminate the need for this board to make changes to the core `esp32/manifest.json`. My changes are across files so I can't share a patch for you to try it now. We'll hold off on merging this until that's available to update against.

stc1988:
I migrated m5nanoh2 to ECMA-419.
I also reverted the changes to esp32/manifest.json.

phoddie:
Thank you! That matches the new patterns nicely. We'll get this merged.

mkellner:
Merged. It will be available in the next release.

PR #1677, opened 2026-08-11
add tm1637 7-segment LED display

stc1988:
I confirmed with [Seed 4-Digit Display](https://wiki.seeedstudio.com/ja/Grove-4-Digit_Display/).

### DIFF
--- a/modules/drivers/peripherals/tm1637/tm1637.js
+++ b/modules/drivers/peripherals/tm1637/tm1637.js
@@ -1,0 +1,350 @@
+/*
+ * Copyright (c) 2026  Moddable Tech, Inc.
+ *
+ *   This file is part of the Moddable SDK Runtime.
+ *
+ *   The Moddable SDK Runtime is free software: you can redistribute it and/or modify
+ *   it under the terms of the GNU Lesser General Public License as published by
+ *   the Free Software Foundation, either version 3 of the License, or
+ *   (at your option) any later version.
+ *
+ *   The Moddable SDK Runtime is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU Lesser General Public License for more details.
+ *
+ *   You should have received a copy of the GNU Lesser General Public License
+ *   along with the Moddable SDK Runtime.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/*
+	Titan Micro Electronics TM1637 LED display driver
+
+	Supports displays with up to 6 digits. Most commonly available modules have
+	4 digits, with the colon connected to bit 7 of the second digit.
+
+	write(string) displays the string. When the string is longer than the
+	available digits it automatically scrolls at a configurable rate
+	(default 250 ms per step).
+*/
+
+import Timer from "timer";
+
+const CMD_DATA = 0x40;
+const CMD_ADDRESS = 0xC0;
+const CMD_DISPLAY = 0x80;
+const DISPLAY_ON = 0x08;
+const MAX_DIGITS = 6;
+
+// 7-segment font. Bit order: dp g f e d c b a (bit 0 = segment A)
+const FONT = Object.freeze({
+	0x20: 0b00000000, // space
+	0x2D: 0b01000000, // -
+	0x30: 0b00111111, // 0
+	0x31: 0b00000110, // 1
+	0x32: 0b01011011, // 2
+	0x33: 0b01001111, // 3
+	0x34: 0b01100110, // 4
+	0x35: 0b01101101, // 5
+	0x36: 0b01111101, // 6
+	0x37: 0b00000111, // 7
+	0x38: 0b01111111, // 8
+	0x39: 0b01101111, // 9
+	0x3D: 0b01001000, // =
+	0x41: 0b01110111, // A
+	0x42: 0b01111100, // B
+	0x43: 0b00111001, // C
+	0x44: 0b01011110, // D
+	0x45: 0b01111001, // E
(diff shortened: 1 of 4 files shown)


stc1988:
https://github.com/user-attachments/assets/8f095643-7349-40c9-acca-db10f5b89326

modules/drivers/peripherals/tm1637/tm1637.js
```diff
@@ -0,0 +1,350 @@
...
+	 * Write a string to the display. If the string fits it is shown statically
+	 * (right-padded with spaces). Otherwise continuous scrolling begins.
+	 *
+	 * The characters '.' and ':' light bit 7 of the preceding digit and do not
+	 * consume a digit position. On common clock modules, "12:34" lights the
+	 * colon because bit 7 of the second digit is connected to it.
+	 */
+	write(string) {
+		if (typeof string !== "string")
+			string = String(string);
```
phoddie: The typeof check isn't necessary - faster and smaller to unconditionally call `String()`

modules/drivers/peripherals/tm1637/tm1637.js
```diff
@@ -0,0 +1,350 @@
...
+			const cycleLength = this.#content.length + this.#digitCount;
+			this.#offset = (this.#offset + this.#direction) % cycleLength;
+			if (this.#offset < 0)
+				this.#offset += cycleLength;
+			this.#render();
+		}, this.#rate);
+	}
+
+	#stopScroll() {
+		if (this.#timer) {
```
phoddie: It is safe to call `Timer.clear()` on `undefined`, so this `if` is unnecessary

modules/drivers/peripherals/tm1637/tm1637.js
```diff
@@ -0,0 +1,350 @@
...
+				this.#writeByte(CMD_ADDRESS);
+				for (let i = 0; i < this.#ram.length; i++)
+					this.#writeByte(this.#ram[i]);
+			}
+			finally {
+				this.#stop();
+			}
+		}
+		catch (e) {
+			this.#onError?.(e);
```
phoddie: technically this is not ECMA-419 style because`onError()` can be invoked from the public `write()` API. You can just throw here, since `write()` throwing is normal error behavior. But from the scrolling callback, you cannot throw since there is no one to catch, so your only options is to call `onError()`. You could do that in your timer callback, by catching errors thrown by `#render()` and rerouting them to `#onError()`.

modules/drivers/peripherals/tm1637/tm1637.js
```diff
@@ -0,0 +1,350 @@
...
+	 *  direction: 1 - scroll direction (-1 scroll right)
+	 *  brightness: 7 - display brightness (0–7)
+	 */
+	constructor(options) {
+		const sensor = options?.sensor;
+		if (!sensor?.io || (undefined === sensor.clock) || (undefined === sensor.data))
+			throw new Error("sensor with io, clock, and data is required");
+
+		const digits = options.digits ?? 4;
+		if (!Number.isInteger(digits) || (digits < 1) || (digits > MAX_DIGITS))
```
phoddie: JavaScript builtins that accept a number for an integer generally round the number rather than throwing if there is a fractional part. Maybe change to `const digits = Math.round(options.digits) ?? 4` instead.

modules/drivers/peripherals/tm1637/tm1637.js
```diff
@@ -0,0 +1,350 @@
...
+		if (undefined !== options.brightness)
+			this.#brightness = Math.max(0, Math.min(7, options.brightness | 0));
+
+		try {
+			const {io: Digital, clock, data, ...ioOptions} = sensor;
+			this.#clock = new Digital({
+				...ioOptions,
+				pin: clock,
+				mode: Digital.OutputOpenDrain,
+				initialValue: 1
```
phoddie, line 121: Nice use of new `initialValue` property!

modules/drivers/peripherals/tm1637/tm1637.js
```diff
@@ -0,0 +1,350 @@
...
+		if (!Number.isInteger(digits) || (digits < 1) || (digits > MAX_DIGITS))
+			throw new RangeError("digits must be an integer from 1 to 6");
+
+		this.#onError = options.onError;
+		if (undefined !== options.rate)
+			this.#rate = Math.max(50, options.rate | 0);
+		if (undefined !== options.direction)
+			this.#direction = (options.direction === -1) ? -1 : 1;
+		if (undefined !== options.brightness)
+			this.#brightness = Math.max(0, Math.min(7, options.brightness | 0));
```
phoddie: If you move this down further in constructor, you might just be able to do `this.configure(options);` though... arguably these should not be options to the constructor, and the app should call `configure()` separately. (This [guide](https://www.moddable.com/guides/implement419/configure) hints at that, but as an implementation guide, rather than an API design guide, it doesn't go into details).

phoddie (commented):
Overall this looks really good and follows the ECMA-419 conventions well.

I added some comments for your consideration.

stc1988:
Thank you for the review. I had AI check whether the implementation followed the Guide, but it seems to have missed it. As for `onError`, the same can be said for [qwiic_alphanumeric.js](https://github.com/Moddable-OpenSource/moddable/blob/b1f42a2e148f0fc2cd91d7ed1cee56bd361656b2/modules/drivers/peripherals/qwiic_alphanumeric/qwiic_alphanumeric.js#L142),  which I used as a reference.

I have addressed the points you pointed out, so could you please check them?

phoddie:
Thank you for the changes. It looks good now.

We'll look at updating the qwiic peripheral you modeled this on. You demonstrate why correct code, not just working code, is important: people (and their LLMs) assume existing code is a good model. But, neither checks the documentation. ;)

mkellner:

PR #1681, opened 2026-08-19
CoreS3: keep I2S speaker and microphone half-duplex

meganetaaan:
## Summary

Keep the M5Stack CoreS3 speaker and microphone half-duplex on a single I2S bus. AudioOut now uses the manifest I2S port, no longer drives ES7210's MCLK, and AudioIn no longer enables RX before `start()`.

## Problem

CoreS3 shares BCLK/WS between AW88298 and ES7210. GPIO0 MCLK is wired only to ES7210.

The generic ESP32 AudioOut constructor always called `I2S_CHANNEL_DEFAULT_CONFIG(I2S_NUM_AUTO, ...)`, so it ignored `audioOut.i2s.num`. AudioIn enabled the RX channel as soon as the worker task started, even while the application still considered the microphone idle. CoreS3 also configured AudioOut with `mck_pin: 0`.

Together, a 24 kHz speaker turn can leave the shared clock and codec in a state that corrupts the next 16 kHz microphone turn.

## Changes

- `audioout-esp32.c`: use `MODDEF_AUDIOOUT_I2S_NUM` (default `I2S_NUM_AUTO`) and IDF Philips/MSB slot macros.
- CoreS3 `audioOut.i2s`: keep I2S1, add `format_i2s: 1`, remove speaker `mck_pin`.
- `audioin.c`: create the standard RX channel in READY from the constructor, enable it only after `start()`, initialize `bytes_read`, and ignore timeout/unaligned reads.
- CoreS3 wrappers take a single bus owner and call `amp.start`/`amp.stop` and `mic.start`/`mic.stop` when the half-duplex role changes.

This does not add a shared TX/RX duplex controller. 16 kHz capture and 24 kHz playback still alternate.

## Validation

- The constructor-time AudioIn setup and AudioOut I2S port selection were reviewed against the ESP-IDF standard I2S APIs.
- A CoreS3 firmware build/flash cycle was **not** run in this environment.

## Notes

Projects that remap `embedded:io/audio/in` back to the generic ESP32 module will still get the C-level READY/`start()` change, but not the CoreS3 codec owner.

### DIFF
--- a/build/devices/esp32/targets/m5stack_cores3/M5StackCoreS3AudioIn.js
+++ b/build/devices/esp32/targets/m5stack_cores3/M5StackCoreS3AudioIn.js
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Shinya Ishikawa
+ * Copyright (c) 2026 Shinya Ishikawa
  *
  *   This file is part of the Moddable SDK Runtime.
  *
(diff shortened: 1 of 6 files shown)


meganetaaan:
Closing this for now. I opened it before doing comprehensive hardware validation, which was premature.

The AudioIn changes affect when the RX channel is created and enabled, so I want to test that path carefully on a CoreS3 and other devices before asking for review. I will reopen a PR after that validation.

PR #1683, opened 2026-08-24
Align HTTP request write typings with ECMA-419

stc1988:
ECMA-419 specifies that calling write() with no arguments signals the end of a request body using chunked transfer encoding. It also specifies that write() returns the number of bytes that may subsequently be written.


phoddie:
Absolutely correct. We'll get this merged. Thank you.

mkellner:

PR #1684, opened 2026-08-25
Update mcp230xx.js constructor

Templarian:
If one sets any other config it will overwrite the default address.


phoddie:
> If one sets any other config it will overwrite the default address.

Very true.

We'll get this merged for the September release. Thank you.

mkellner:

PR #1689, opened 2026-08-26
Update expander.js 0x06 should be 0x0C for pullups

Templarian:
I believe in both places it should be `0x0C` (as `0x06` is DEFVAL).

It works either way with my mcp23017 for interrupts on buttons. Might be why it wasn't noticed.

Still new to this chip so could be reading the library wrong.


phoddie:
It took me a few minutes to remember how this works. ;)

It looks like you are correct. I think what happened is that it used the Bank = 1 mode register for GPPUA (0x06) but since the expander runs in Bank = 0 mode, GPPUA is different (0x0C).

Thank you! We'll get this merged for the September release.

mkellner:

PR #1693, opened 2026-09-07
Fix listen query forwarding and URL searchParams synchronization

### LINKED ISSUE #1692: Query parameters are lost in the listen example
**Build environment:** macOS
**Moddable SDK version:** 9.5.0
**Target device:** desktop simulator

**Description**

In `examples/io/listener/listen`, query parameters are missing from request.url.

**Steps to Reproduce**
1. run `examples/io/listener/listen`
2. send request `curl 'http://localhost/query?text=hello'`

**Expected behavior**
request.url.searchParams.get("text") returns "hello".
Actual: It returns null, and request.url.search is empty.

**Other information**

The HTTP server separates the request target into path and query, but listen.js constructs the URL using only path:

https://github.com/Moddable-OpenSource/moddable/blob/public/examples/io/listener/listen/listen.js#L154-L155

stc1988:
Title: Fix query parameters in listen and URL searchParams synchronization

Fixes #1692.

Preserve query parameters when constructing request URLs in `listen`. Fix `URL.searchParams` synchronization when setting `search` on a URL without a query or replacing `href`, while preserving the existing `searchParams` object.

### DIFF
--- a/examples/io/listener/listen/listen.js
+++ b/examples/io/listener/listen/listen.js
@@ -152,7 +152,9 @@ async function* listen(options) {
 					_connection_.accept({
 						onRequest(_request_) {
 							const { method, path, headers } = _request_;
-							const request = new Request(new URL(path, base), { method, path, headers, body:requestPromise });
+							const url = new URL(path, base);
+							url.search = _request_.query;
+							const request = new Request(url, { method, path, headers, body:requestPromise });
 							const connection = {
 								close() {
 									_connection_.close();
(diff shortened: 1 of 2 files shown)


mkellner:

phoddie:
Thank you for the PR.

It would be helpful if future PRs could include code fragments to demonstrate the problem, not only a written summary. In this case, one way to do that is to extend the [unit tests](https://github.com/Moddable-OpenSource/moddable/tree/public/tests/modules/data/url) to cover the changes. We've update the unit tests as part of merging the PR.

PR #1694, opened 2026-09-12
Listener construct failure on lin/mac omits xsForget

cmidgley:
Moddable: 9.5.0
Platform: Linux and Mac

On Linux and Mac (not Windows) simulators, `xs_listener_constructor` calls `xsRemember` then, on bind/listen failure, frees the host data without `xsForget`. That leaves a dangling remembered root; after repeated failures and GC the VM can abort. 

The following will reproduce the failure:
- [main.js](https://github.com/user-attachments/files/32150992/main.js)
- [manifest.json](https://github.com/user-attachments/files/32150993/manifest.json)

Note: I do not currently have a working Mac  setup, so I've only verified the Linux patch.

### DIFF
--- a/modules/io/socket/lin/tcp.c
+++ b/modules/io/socket/lin/tcp.c
@@ -558,7 +558,7 @@ static const xsHostHooks xsListenerHooks = {
 
 void xs_listener_constructor(xsMachine *the)
 {
-	Listener listener;
+	Listener listener = C_NULL;
 	int port = 0;
 	xsSlot *onReadable;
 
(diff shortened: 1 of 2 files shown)


phoddie:
All true.

FWIW:

- Windows should also have the check for `c_calloc` failing. We can take care of that after merging.
- No harm in calling `xsmcSetHostData(xsThis, NULL);` or  `xs_listener_destructor_(listener);` in `xsCatch` block (as Windows does), even when listener is NULL.

mkellner:

PR #1695, opened 2026-09-12
WebSocketClient masking issues (on TCP split frames and control frames)

cmidgley:
Moddable: 9.5.0
Platform: all (verified only on Linux simulator and ESP32)

In `embedded:network/websocket/client`, I hit two masking bugs:

1. When a masked data frame is split across TCP reads, `read()` unmasks the first chunk but skips mask rotation when that read drains `#data` (even though more of the same frame remains).
2. Control-frame payloads (close/ping) are read without unmasking, so close codes delivered to `onControl` are garbage.

The following will reproduce the failure:
- [main.js](https://github.com/user-attachments/files/32151305/main.js)
- [manifest.json](https://github.com/user-attachments/files/32151306/manifest.json)

### DIFF
--- a/examples/io/tcp/websocketclient/websocketclient.js
+++ b/examples/io/tcp/websocketclient/websocketclient.js
@@ -237,18 +237,16 @@ class WebSocketClient {
 		if (options.mask) {
 			const mask = options.mask;
 			Logical.xor(data, mask, count);
-			if (this.#data) {
-				switch (count & 3) {
-					case 1:
-						options.mask = Uint8Array.of(mask[1], mask[2], mask[3], mask[0]);
-						break;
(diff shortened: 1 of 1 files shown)


phoddie:
Hey, @cmidgley! I always knew we couldn't consider the WebSocket implementation complete until you had a chance to take a good look at it. ;)

The changes make sense. We'll get them merged.

FWIW – We use the WebSocketClient extensively (ChatAudioIO uses it to stream ridiculous amounts of data) and it's very reliable. These issues are a consequence of reversing the direction and using the client for a server, which beats up unmasking support. That's why we didn't notice.

mkellner:

